Issue #263, opened 2021-11-03
faster nms implementation


fanweiya:
I used sahi to cut the image and count the cells.this is [image.](https://kdocs.cn/l/sak4HJobioT8),But sahi spent a lot of time in postprocessing, and this kind of image took about ten Minutes.

this my code.
```
from sahi.model import Yolov5DetectionModel
from sahi.utils.cv import read_image
from sahi.utils.file import download_from_url
from sahi.predict import get_prediction, get_sliced_prediction, predict
from imutils.paths import list_images
imgpath="test.jpg"
detection_model = Yolov5DetectionModel(
    model_path="best.pt",
    confidence_threshold=0.1,
    device="cuda", # or 'cuda'
)
result = get_sliced_prediction(
    imgpath,
    detection_model,
    slice_height=1024,
    slice_width=1024,
    overlap_height_ratio=0.2,
    overlap_width_ratio=0.2,
)
result.export_visuals(export_dir=r"H:\test")
```

fcakyon:
@fanweiya when the number of proposals is very large, it is expected from postprocess to take long. You can increase the confidence threshold for faster postprocessing.

fanweiya:
you can use https://github.com/remydubois/lsnms

fcakyon:
@fanweiya thanks for your feedback! It is now in our roadmap but no ETA. We are open to any PR!

phborba:
Hi, @fcakyon, how are you doing? First of all, congrats on the framework! It is fantastic! Are you guys interested in putting the lsnms that @fanweiya suggested? If you guys are willing to integrate it, maybe I could do it for you. What do you think?

fcakyon:
@phborba i am happy to hear your nice words on the framework!

If you are willing to implement lsnms into `sahi`, we are open to accept your PR!

What you have to do is to add a new class named `LSNMSPostprocess` similar to our current [NMSPostprocess class](https://github.com/obss/sahi/blob/21ecb285aa6bf93c2a00605dfb9b138f19d8d62d/sahi/postprocess/combine.py#L103).

remydubois:
Hey, I am the developer of the `lsnms` library.
If you guys @phborba @fcakyon need help in integrating `lsnms.nms` into `sahi`, feel free to tag me in this thread.  

Cheers

fanweiya:
lsnms is fast, but How do I quickly make Numba warmup?

remydubois:
[...]

PR #301, opened 2021-12-09
add lsnms postprocess

fcakyon:
solves: https://github.com/obss/sahi/issues/263

based on: https://github.com/remydubois/lsnms

### DIFF
--- a/sahi/postprocess/combine.py
+++ b/sahi/postprocess/combine.py
@@ -221,7 +221,9 @@ class LSNMSPostprocess(PostprocessPredictions):
         try:
             from lsnms import nms
         except ModuleNotFoundError:
-            raise ModuleNotFoundError('Please run "pip install -U lsnms" to install lsnms first for lsnms utilities.')
+            raise ModuleNotFoundError(
+                'Please run "pip install lsnms>0.3.1" to install lsnms first for lsnms utilities.'
+            )
 
         class_ids = np.array([object_prediction.category.id for object_prediction in object_predictions])
         boxes = np.array([object_prediction.bbox.to_voc_bbox() for object_prediction in object_predictions])
--- a/sahi/postprocess/combine.py
+++ b/sahi/postprocess/combine.py
@@ -211,7 +211,7 @@ class UnionMergePostprocess(PostprocessPredictions):
             shift_amount=mask1.shift_amount,
         )
 
-    
+
 class LSNMSPostprocess(PostprocessPredictions):
     # https://github.com/remydubois/lsnms/blob/10b8165893db5bfea4a7cb23e268a502b35883cf/lsnms/nms.py#L62
     def __call__(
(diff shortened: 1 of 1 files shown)


fcakyon:
Can you please check if my implementation is correct? :) @remydubois @phborba

remydubois:
Hey, I think everything is correct for the `nms` part 🎉 

However, weighted box clustering (`wbc`) returns a new set of boxes and scores (because it "creates" instances that did not exist in the input list: by averaging boxes and scores cluster-wise).  
I think it would look like something like this:
```python3
boxes, scores, _ = wbc(boxes, scores, self.match_threshold)
```
Having said that, I am not really sure how WBC should be applied in a multi class setting: if a car and a horse overlap, it does not make much sense to me to aggregate them together... what do you think @fcakyon ?
(note that the problem does not arise with NMS since we just discard the lowest scoring one).

fcakyon:
You are right integrating wbc into sahi is not trivial. Will remove it for now 👍

remydubois:
okay :) LGTM then

fcakyon:
@remydubois i have one last question. Current implementation applies class agnostic nms, right? (so it can combine predictions from different categories)

remydubois:
@fcakyon  Absolutely, it is completely class agnostic. If a multi class NMS seems useful to you I could try and see how to integrate it while keeping performance level high.

On a different topic: I know that at first use, NMS can lag a bit (10-15 seconds), which is due to Numba's inner compiling of recursive functions. I know that they have been trying to solve the issue ([cf this issue](https://github.com/numba/numba/issues/6713#issue-804901574)) but it is still pending as of now.  
I will update `lsnms` as soon as this is fixed (hopefully it will one day).

fcakyon:
> @fcakyon Absolutely, it is completely class agnostic. If a multi class NMS seems useful to you I could try and see how to integrate it while keeping performance level high.

High performance multi class NMS implementation would be awesome!

remydubois:
I will release it version 0.3.0 soon, the api should change a little bit so perhaps wait that i release it, and adapt this branch before merging it
(Should just be an extra argument ´class_ids’ to lsnms.nms)

fcakyon:
Awesome news @remydubois!

remydubois:
@fcakyon `lsnms` version 0.3.X should be published on pypi and handle the multiclass setting, with a minimal modification of the API.

fcakyon:
@remydubois i am trying to add lsnms support with the next release but having this error when testing: 

```
Exception has occurred: ValueError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
Class ids should be a positive integer array. Received float32 with min 0.0
  File "/home/xxx/miniconda3/envs/xxx/lib/python3.8/site-packages/lsnms/util.py", line 421, in check_correct_arrays
    raise ValueError(
  File "/home/xxx/miniconda3/envs/xxx/lib/python3.8/site-packages/lsnms/util.py", line 465, in check_correct_input
    check_correct_arrays(boxes, scores, class_ids)
  File "/home/xxx/miniconda3/envs/xxx/lib/python3.8/site-packages/lsnms/nms.py", line 144, in nms
    boxes, scores, class_ids = check_correct_input(
```

In most of the datasets class ids are starting from 0 so i cannot change the class ids..

remydubois:
Hey,
I think the error is fired because you provided class ids as a floating point vector.
The underlying code uses np.can_cast, so if the error is fired, it means the class_ids can not be casted to integer.

Try converting it to integer on your side (or ensure that all the values provided are, at least, round floats).

fcakyon:
@remydubois thanks a lot for the feedback! Will try to cast them to int before calling lsnms 👍

PR #320, opened 2021-12-27
refactor postprocessing and coco eval

fcakyon:
- all postprocess operations migrated to torch: now 100x faster
- refactorized coco eval and error analysis to handle missing categories in dataset

todo: lsnms fails with: ```Class ids should be a positive integer array. Received float32 with min 0.0``` @remydubois


README.md
```diff
@@ -168,7 +168,7 @@ You can specify sliced inference parameters as:
 sahi predict --slice_width 256 --slice_height 256 --overlap_height_ratio 0.1 --overlap_width_ratio 0.1 --model_confidence_threshold 0.25 --source image/file/or/folder --model_path path/to/model --model_config_path path/to/config
 ```
 
-- Specify postprocess type as `--postprocess_type UNIONMERGE` or `--postprocess_type NMS` to be applied over sliced predictions
+- Specify postprocess type as `--postprocess_type GREEDYNMM` or `--postprocess_type NMS` to be applied over sliced predictions
```
devrimcavusoglu, line 171: Maybe we call it `GMS` ? In the paper they used the term Greedy Merge Strategy, we can state a reference for GMS as well. Wdyt ? @fcakyon

sahi/postprocess/legacy/combine.py
```diff
@@ -0,0 +1,213 @@
...
+def calculate_intersection_area(box1: np.ndarray, box2: np.ndarray) -> float:
+    """
+    Args:
+        box1 (np.ndarray): np.array([x1, y1, x2, y2])
+        box2 (np.ndarray): np.array([x1, y1, x2, y2])
+    """
+    left_top = np.maximum(box1[:2], box2[:2])
+    right_bottom = np.minimum(box1[2:], box2[2:])
+    width_height = (right_bottom - left_top).clip(min=0)
+    return width_height[0] * width_height[1]
```
devrimcavusoglu, line 30: I think we can change here like this for better readability.

```python
width, height = (right_bottom - left_top).clip(min=0)
return width * height
````

devrimcavusoglu: One more thing, these helper functions also exist on `sahi/postprocess/combine.py`, to avoid duplication we can move these to some util script. Wdyt ? @fcakyon

README.md
fcakyon: @devrimcavusoglu which paper do you mention about? can you give url?

sahi/postprocess/legacy/combine.py
fcakyon: that would make sense 👍

README.md
devrimcavusoglu: [OverFeat](https://arxiv.org/abs/1312.6229) paper, afaik they introduced this merging strategy first in their paper (as a paper though earlier works couldve been done possibly). I don't know if later the name changed to something more common.

fcakyon: in that paper, they take the average of bbox coordinates when merging, we are taking the union of the predictions. 

Moreover, they merge only bounding boxes while we also merge the predicted segmentation masks.

I think we can give a different name since its not the same technique.

devrimcavusoglu: Oh I mistaken our method as merging by averaging, just like the paper. NMM sounds good now :smile: :+1:

PR #321, opened 2021-12-30
Add export_file_name and export_format argument to export_visuals

Fraccalo:
Add functionality to edit file name when exporting an annotated image with the export_visuals method.


fcakyon:
@Fraccalo thanks a lot for your contributions! Can you please reformat your code following the steps given in: https://github.com/obss/sahi#contributing

After that we are good to merge your PR!

PR #374, opened 2022-02-04
fix annotation dict without segmentation

tureckova:
While slicing the dataset with only bbox annotations, a key error occurred in file sahi/utils/coco.py on line 130.
I fixed the if statement for annotation dict without segmentation.

### DIFF
--- a/sahi/utils/coco.py
+++ b/sahi/utils/coco.py
@@ -127,7 +127,7 @@ class CocoAnnotation:
             annotation_dict: dict
                 COCO formatted annotation dict (with fields "bbox", "segmentation", "category_id")
         """
-        if annotation_dict["segmentation"] and not isinstance(annotation_dict["segmentation"], list):
+        if annotation_dict.__contains__("segmentation") and not isinstance(annotation_dict["segmentation"], list):
             has_rle_segmentation = True
             logger.warning(
                 f"Segmentation annotation for id {annotation_dict['id']} is skipped since RLE segmentation format is not supported."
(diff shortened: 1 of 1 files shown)


fcakyon:
Thanks @tureckova ! Will merge it once the tests pass 👍

sahi/utils/coco.py
```diff
@@ -127,15 +127,15 @@ def from_coco_annotation_dict(cls, annotation_dict: Dict, category_name: Optiona
             annotation_dict: dict
                 COCO formatted annotation dict (with fields "bbox", "segmentation", "category_id")
         """
-        if annotation_dict["segmentation"] and not isinstance(annotation_dict["segmentation"], list):
+        if annotation_dict.__contains__("segmentation") and not isinstance(annotation_dict["segmentation"], list):
```
fcakyon, line 130: here we have to check if `segmentation` field is not empty (object detection datasets in COCO format have `segmentation: []` in the json file)

sahi/utils/coco.py
```diff
@@ -127,15 +127,15 @@ def from_coco_annotation_dict(cls, annotation_dict: Dict, category_name: Optiona
...
             has_rle_segmentation = True
             logger.warning(
                 f"Segmentation annotation for id {annotation_dict['id']} is skipped since RLE segmentation format is not supported."
             )
         else:
             has_rle_segmentation = False
 
-        if annotation_dict["segmentation"] and not has_rle_segmentation:
+        if annotation_dict.__contains__("segmentation") and not has_rle_segmentation:
```
fcakyon, line 138: similarly,

here we have to check if `segmentation` field is not empty (object detection datasets in COCO format have `segmentation: []` in the json file)

tureckova: I thought that only there is the check necessary

fcakyon:
@tureckova thanks a lot for your contributions!

PR #435, opened 2022-04-08
Update inference_for_mmdetection.ipynb

weypro:
Resolve the compatibility error `AttributeError: module 'PIL.Image' has no attribute 'Resampling'`


fcakyon:
@weypro thanks a lot for the PR! Can you briefly mention what you have changed?

weypro:
Only the first code block.
Before:
```
!pip install -U torch sahi mmdet mmcv-full
```
After:
```
!pip install -U pillow==9.0.1 torch==1.10.0 sahi imgaug==0.2.5
```

fcakyon:
@weypro why do we need to fix to an older version of `Pillow`? Is it caused by mmdet/torch or is there anything we can do in `sahi` to support latest `Pillow` version?

weypro:
The error occurs `AttributeError: module 'PIL.Image' has no attribute 'Resampling'` and the version of `Pillow` in colab runtime is 7.1.2. , so I try to fix it by specifying a newer version.
But after further testing, I find the truth that `!pip install -U torch sahi mmdet mmcv-full` was excuted unsuccessfully at that time, which actually contains the latest update of `Pillow`. The error is  a misunderstanding and it's unnecessary to specify the version explicitly, and I will close the PR. Thank you for the reminding.

PR #442, opened 2022-04-10
add support for video prediction

madenburak:
This PR brings full support for sliced/standard inference for video input format for all MMDetection/Detectron2/YOLOv5 models :fire: 

- Just give video file as source:

```bash
sahi predict --model_path yolov5s.pt --model_type yolov5 --source video.mp4 --export_visual
``` 

You can also view video render during video inference with `--view_video`:

```bash
sahi predict --model_path yolov5s.pt --model_type yolov5 --source video.mp4 --export_visual --view_video
``` 

- To `forward 100 frames`, on opened window press key `D `
- To `revert 100 frames`, on opened window press key `A`
- To `forward 20 frames`, on opened window press key `G`
- To `revert 20 frames`, on opened window press key `F`
- To `exit`, on opened window press key `Esc`

Note: If `--view_video` is slow, you can add `--frame_skip_interval=20` argument to skip interval of 20 frames each time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,37 +154,31 @@ Find detailed info on COCO utilities (yolov5 conversion, slicing, subsampling, f
 
 Find detailed info on MOT utilities (ground truth dataset creation, exporting tracker metrics in mot challenge format) at [mot.md](docs/mot.md).
 
+
 ### Input, Export Video Support 
 
-<<<<<<< HEAD
 A video can be given as input and the results can be exported as video. You can also see the SAHI predictions during inference on the interactive window where fast-forwarding or backwarding is also possible.
-=======
-As the input, the video can be given and the results can be exporting as video. It can also see the SAHI prediction result during inference on opened window, video can be skip or prev. However, it can be still predicted from the image.
->>>>>>> 100f6fdbdafae3ee67b2bd825c7c92aa0dadb4db
 
 One more thing added prediction time to terminal.
 
-For input, export video:
-` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4`   
+#### For input, export video:
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source home/user/video.mp4 --input_video --export_visual`   
 
-For view result of prediction during inference:
-` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4  --view_img` 
+#### For view result of prediction during inference:
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --input_video --export_visual --view_image `
 
-<<<<<<< HEAD
-For fast-forwarding or backwarding at result view of prediction during inference::
-=======
-For skip or prev view result of prediction during inference:
->>>>>>> 100f6fdbdafae3ee67b2bd825c7c92aa0dadb4db
-` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path /home/burak/video.mp4  --view_img --forward_backward` 
+#### For fast-forwarding at result view of prediction during inference:
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --input_video --export_visual --view_image --fast_forwarding` 
 
 * Skip 100 frames, on opened window  press key = d 
 * Prev 100 frames, on opened window press key = a
-* Skip 5 frames, on opened window press key = g
-* Prev 5 frames, on opened window press key = f
+* Skip 10 frames, on opened window press key = g
+* Prev 10 frames, on opened window press key = f
 * Exit,  on opened window press key = Esc
 
-For input, export image:
-` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/burak/picture.png`
+#### For input, export image:
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/picture.png --export_visual`
+
 
 ## <div align="center">Citation</div>
 
--- a/README.md
+++ b/README.md
@@ -165,16 +165,15 @@ One more thing added prediction time to terminal.
 `sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source home/user/video.mp4 --export_visual`   
 
 #### For view result of prediction during inference:
-`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --export_visual --view_image `
-
-#### For fast-forwarding at result view of prediction during inference:
-`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --export_visual --view_image --fast_forwarding` 
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --view_image`
 
 * Skip 100 frames, on opened window  press key = d 
 * Prev 100 frames, on opened window press key = a
-* Skip 10 frames, on opened window press key = g
-* Prev 10 frames, on opened window press key = f
+* Skip 20 frames, on opened window press key = g
+* Prev 20 frames, on opened window press key = f
 * Exit,  on opened window press key = Esc
+
+Note: If view_image is slow, you can add `--time_interval=(your choice)` argument without brackets as int. If you want to export video during view_image, in addition you must use `--export_visual` and `--avarage_prediction_time=(your choice)`. Details in predict.py file.
 
 #### For input, export image:
 `sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/picture.png --export_visual`
--- a/README.md
+++ b/README.md
@@ -154,6 +154,30 @@ Find detailed info on COCO utilities (yolov5 conversion, slicing, subsampling, f
 
 Find detailed info on MOT utilities (ground truth dataset creation, exporting tracker metrics in mot challenge format) at [mot.md](docs/mot.md).
 
+### Input, Export Video Support 
+
+A video can be given as input and the results can be exported as video. You can also see the SAHI predictions during inference on the interactive window where fast-forwarding or backwarding is also possible.
+
+One more thing added prediction time to terminal.
+
+For input, export video:
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4`   
+
+For view result of prediction during inference:
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4  --view_img` 
+
+For fast-forwarding or backwarding at result view of prediction during inference::
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path /home/burak/video.mp4  --view_img --forward_backward` 
+
+* Skip 100 frames, on opened window  press key = d 
+* Prev 100 frames, on opened window press key = a
+* Skip 5 frames, on opened window press key = g
+* Prev 5 frames, on opened window press key = f
+* Exit,  on opened window press key = Esc
+
+For input, export image:
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/burak/picture.png`
+
 ## <div align="center">Citation</div>
 
 If you use this package in your work, please cite it as:
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ One more thing added prediction time to terminal.
 * Prev 20 frames, on opened window press key = f
 * Exit,  on opened window press key = Esc
 
-Note: If view_image is slow, you can add `--time_interval=(your choice)` argument without brackets as int. If you want to export video during view_image, in addition you must use `--export_visual` and `--avarage_prediction_time=(your choice)`. Details in predict.py file.
+Note: If view_image is slow, you can add `--time_interval=(your choice)` argument without brackets as int. If you want to export video during view_image, in addition you must use `--export_visual`. Details in predict.py file.
 
 #### For input, export image:
 `sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/picture.png --export_visual`
--- a/README.md
+++ b/README.md
@@ -162,13 +162,13 @@ A video can be given as input and the results can be exported as video. You can 
 One more thing added prediction time to terminal.
 
 #### For input, export video:
-`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source home/user/video.mp4 --input_video --export_visual`   
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source home/user/video.mp4 --export_visual`   
 
 #### For view result of prediction during inference:
-`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --input_video --export_visual --view_image `
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --export_visual --view_image `
 
 #### For fast-forwarding at result view of prediction during inference:
-`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --input_video --export_visual --view_image --fast_forwarding` 
+`sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --source /home/user/video.mp4 --export_visual --view_image --fast_forwarding` 
 
 * Skip 100 frames, on opened window  press key = d 
 * Prev 100 frames, on opened window press key = a
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Object detection and instance segmentation are by far the most important fields 
 
 | Command  | Description  |
 |---|---|
-| [predict](https://github.com/obss/sahi/blob/main/docs/cli.md#predict-command-usage)  | perform sliced/standard prediction using any [yolov5](https://github.com/ultralytics/yolov5)/[mmdet](https://github.com/open-mmlab/mmdetection)/[detectron2](https://github.com/facebookresearch/detectron2) model |
+| [predict](https://github.com/obss/sahi/blob/main/docs/cli.md#predict-command-usage)  | perform sliced/standard video/image prediction using any [yolov5](https://github.com/ultralytics/yolov5)/[mmdet](https://github.com/open-mmlab/mmdetection)/[detectron2](https://github.com/facebookresearch/detectron2) model |
 | [predict-fiftyone](https://github.com/obss/sahi/blob/main/docs/cli.md#predict-fiftyone-command-usage)  | perform sliced/standard prediction using any [yolov5](https://github.com/ultralytics/yolov5)/[mmdet](https://github.com/open-mmlab/mmdetection)/[detectron2](https://github.com/facebookresearch/detectron2) model and explore results in [fiftyone app](https://github.com/voxel51/fiftyone) |
 | [coco slice](https://github.com/obss/sahi/blob/main/docs/cli.md#coco-slice-command-usage)  | automatically slice COCO annotation and image files |
 | [coco fiftyone](https://github.com/obss/sahi/blob/main/docs/cli.md#coco-fiftyone-command-usage)  | explore multiple prediction results on your COCO dataset with [fiftyone ui](https://github.com/voxel51/fiftyone) ordered by number of misdetections |
@@ -50,6 +50,8 @@ Object detection and instance segmentation are by far the most important fields 
 - [Introduction to SAHI](https://medium.com/codable/sahi-a-vision-library-for-performing-sliced-inference-on-large-images-small-objects-c8b086af3b80)
 
 - [Official paper](https://arxiv.org/abs/2202.06934) (NEW)
+
+- [Video inference support is live](https://github.com/obss/sahi/issues/457) (NEW)
 
 - [Kaggle notebook](https://www.kaggle.com/remekkinas/sahi-slicing-aided-hyper-inference-yv5-and-yx) (NEW)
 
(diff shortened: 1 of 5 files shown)


fcakyon:
Thanks a lot for your contributions @madenburak 🚀 

Tests fail at format check, can you please reformat your code with black&isort as detailed in: https://github.com/obss/sahi#contributing

madenburak:
> Thanks a lot for your contributions @madenburak 🚀
> 
> Tests fail at format check, can you please reformat your code with black&isort as detailed in: https://github.com/obss/sahi#contributing

You're welcome @fcakyon. I can correct this problem on tomorrow night. Also congrats for SAHI, it is proud work.

fcakyon:
@madenburak take your time, no hurries 👍 

Thanks for the kind words, from now on you are also a SAHI contributor 💯

devrimcavusoglu:
@madenburak Thanks for the contribution, I'll review at ASAP.

sahi/predict.py
```diff
@@ -460,14 +472,97 @@ def predict(
...
     durations_in_seconds["slice"] = 0
-    for ind, image_path in enumerate(tqdm(image_path_list, "Performing inference on images")):
+
+    # Input video
+    if export_video:
+        if video_path is not None:
+            #get video name with extension
+            image_base = os.path.basename(video_path)
+        else:
+            print("\nVideo path is none. Enter (--video_path 'path' --export_video) in terminal!\n")
```
devrimcavusoglu, line 482: Shouldn't we raise exception (`ValueError`) here ? If we rely on L484, then just simply remove the `else` block.

**A side note:** `"\n"` are redundant here I think, you do not need these in Python as it will by default print with a new line, and end with a newline.

sahi/predict.py
```diff
@@ -314,7 +318,11 @@ def predict(
     postprocess_match_metric: str = "IOS",
     postprocess_match_threshold: float = 0.5,
     postprocess_class_agnostic: bool = False,
-    export_visual: bool = False,
+    export_visual: bool = True,
+    export_video: bool = False,
+    view_image: bool = False,
+    video_path: str = None,
+    forward_backward: bool = False,
```
devrimcavusoglu, line 325: Provide small descriptions in the docstring for these newly added parameters.

sahi/predict.py
```diff
@@ -460,14 +472,97 @@ def predict(
...
+        # get video from video path
+        cap = cv2.VideoCapture(video_path)
+        # get video properties
+        fps = cap.get(cv2.CAP_PROP_FPS)
+        w = int(cap.get(cv2.CAP_PROP_FRAME_WIDTH))
+        h = int(cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
+        size = (w, h)
+
+        fourcc = cv2.VideoWriter_fourcc(*'mp4v')
+        out = cv2.VideoWriter(str(save_dir) + '/{}'.format(image_base), fourcc, fps, size)
```
devrimcavusoglu, line 492: `save_dir` is already `pathlib.Path` object, you can append the path like, and moreover this type of manual joining as strings can be problematic.

```python
new_path = save_dir / "my_new_path"
```

sahi/predict.py
```diff
@@ -460,14 +472,97 @@ def predict(
...
+        if forward_backward:
+            
+            def get_video(cap):
+                while cap.isOpened:
+                    k = cv2.waitKey(30)
+
+                    frame_num = cap.get(cv2.CAP_PROP_POS_FRAMES)
+
+                    if k == 27:
+                        print('\nQuitting...\n') # Exit the prediction, Key = Esc
```
devrimcavusoglu, line 506: We can remove those `"\n"` (we can refactor those in the whole change), and optionally replace with something else, for example

```python
===== Closing =====
```

sahi/predict.py
```diff
@@ -460,14 +472,97 @@ def predict(
...
+
+        fourcc = cv2.VideoWriter_fourcc(*'mp4v')
+        out = cv2.VideoWriter(str(save_dir) + '/{}'.format(image_base), fourcc, fps, size)
+
+        if view_image:
+            cv2.namedWindow('Prediction of {}'.format(str(image_base)), cv2.WINDOW_AUTOSIZE)
+        
+        if forward_backward:
+            
+            def get_video(cap):
```
devrimcavusoglu, line 499: Rather than defining a sub-function inside `predict`, move this function out (maybe to utils @fcakyon ?) with a parameter `fast_forwarding: bool`, and handle this case in the function.

sahi/predict.py
```diff
@@ -416,6 +424,10 @@ def predict(
 
     # for profiling
     durations_in_seconds = dict()
+    
+    # Need this line, if we don't give images to model 
+    if source is None:
```
devrimcavusoglu, line 429: Can you elaborate more why we need this, isn't [this](https://github.com/obss/sahi/blob/b7fb59d6a41260d8a6318549efcb4d4cb9718bf5/sahi/predict.py#L431) sufficient ?

sahi/predict.py
```diff
@@ -579,9 +676,29 @@ def predict(
...
+            )
+ 
         time_end = time.time() - time_start
         durations_in_seconds["export_files"] = time_end
 
+        if export_video:
+            out.write(get_image)
+        else:
+            Path(output_dir).mkdir(parents=True, exist_ok=True)
+            filename = filename_without_extension = Path(relative_filepath).stem
```
devrimcavusoglu, line 698: We can remove `filename`

sahi/utils/cv.py
```diff
@@ -380,9 +380,9 @@ def visualize_object_predictions(
         Path(output_dir).mkdir(parents=True, exist_ok=True)
         # save inference result
         save_path = os.path.join(output_dir, file_name + "." + export_format)
-        cv2.imwrite(save_path, cv2.cvtColor(image, cv2.COLOR_RGB2BGR))
+        
     elapsed_time = time.time() - elapsed_time
-    return {"image": image, "elapsed_time": elapsed_time}
+    return {"image": image, "elapsed_time": elapsed_time}, image
```
devrimcavusoglu, line 385: Why changing return type from dict to tuple, and we already have `image` in the return dict, why we need an extra ?

devrimcavusoglu (changes requested):

madenburak:
> @madenburak Thanks for the contribution, I'll add my review ASAP.

You're welcome @devrimcavusoglu, I will complete all your reviews.

sahi/predict.py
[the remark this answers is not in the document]
madenburak: Is it fast_forwarding_backwarding, because video can be backwarding.

sahi/predict.py
madenburak: I didn't think this, your opinion is better than

sahi/predict.py
madenburak: Do you want to I use path like: `new_path = save_dir / "my_new_path"` `out = cv2.VideoWriter(new_path, fourcc, fps, size) ` or `new_path = os.path.join(save_dir, image_base`

sahi/predict.py
madenburak: When we remove those "\n",  text is written same line. But we can remove at end of text.

sahi/utils/cv.py
madenburak: You are right. I didn't pay attention to return

sahi/predict.py
madenburak: Did you decide that? @devrimcavusoglu @fcakyon

sahi/predict.py
madenburak: Normally, `source` is used for image inference and this is many parts of code. If we give input video, `source` variable is none. The code give error to us at this situation, so I used it like that. But If we inference on image, source is not none and this not problem. Have you better idea for this? Because my idea is quick solution.

sahi/predict.py
```diff
@@ -314,7 +316,11 @@ def predict(
     postprocess_match_metric: str = "IOS",
     postprocess_match_threshold: float = 0.5,
     postprocess_class_agnostic: bool = False,
-    export_visual: bool = False,
+    export_visual: bool = True,
+    export_video: bool = False,
+    view_image: bool = False,
+    video_path: str = None,
```
fcakyon, line 322: lets remove this var and take the video path with `source` parameter

README.md
```diff
@@ -154,6 +154,33 @@ Find detailed info on COCO utilities (yolov5 conversion, slicing, subsampling, f
...
+For input, export video:
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4`   
+
+For view result of prediction during inference:
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4  --view_img` 
+
+
+For fast-forwarding or backwarding at result view of prediction during inference::
+
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path /home/burak/video.mp4  --view_img --forward_backward`
```
fcakyon, line 173: update the argument `--forward_backward` with `--fast_forward_backward`

sahi/predict.py
fcakyon: @madenburak can we make it like this?
1. perform image inference using `source` when `export_video=False`
2. perform video inference using `source` when `export_video=True`

sahi/predict.py
fcakyon: @madenburak we can create a function in `sahi/utils/ct.py` named `get_video_reader` as:

```python
def get_video_reader(video_path: str, view_image: bool = False, fast_forwarding_backwarding: bool = False):
    # get video name with extension
    image_base = os.path.basename(video_path)
    # get video from video path
    cap = cv2.VideoCapture(video_path)

    if fast_forwarding_backwarding:
        def read_video_frames(cap):
            while cap.isOpened:
                k = cv2.waitKey(30)

                frame_num = cap.get(cv2.CAP_PROP_POS_FRAMES)

                if k == 27:
                    print(
                        "\n===========================Closing==========================="
                    )  # Exit the prediction, Key = Esc
                    exit()
                if k == 100:
                    frame_num += 100  # Skip 100 frames, Key = d
                if k == 97:
                    frame_num -= 100  # Prev 100 frames, Key = a
                if k == 102:
                    frame_num += 5  # Skip 5 frames, Key = g
                if k == 103:
                    frame_num -= 5  # Prev 5 frames, Key = f
                cap.set(cv2.CAP_PROP_POS_FRAMES, frame_num)

                ret, frame = cap.read()
                if not ret:
                    print("\n===========================Video Ended===========================")
                    break
                yield Image.fromarray(frame)

                if view_image:
                    cv2.imshow("Prediction of {}".format(str(image_base)), get_image_from_dict)
                    cv2.waitKey(1)

    else:
        def read_video_frames(cap):
            while cap.isOpened:
                k = cv2.waitKey(30)

                if k == 27:
                    print(
                        "\n===========================Closing==========================="
                    )  # Exit the prediction, Key = Esc
                    exit()

                ret, frame = cap.read()
                if not ret:
                    print("\n===========================Video Ended===========================")
                    break
                yield Image.fromarray(frame)

                if view_image:
                    cv2.imshow("Prediction of {}".format(str(image_base)), get_image_from_dict)
                    cv2.waitKey(1)

    return read_video_frames
```

fcakyon: then you can use this `get_video_reader` in `sahi/predict.py` as:

```python
video_capture, read_video_frame = get_video_reader(video_path, view_image, fast_forwarding_backwarding)
item = read_video_frame(video_capture)
```

sahi/utils/cv.py
```diff
@@ -380,7 +380,7 @@ def visualize_object_predictions(
         Path(output_dir).mkdir(parents=True, exist_ok=True)
         # save inference result
         save_path = os.path.join(output_dir, file_name + "." + export_format)
-        cv2.imwrite(save_path, cv2.cvtColor(image, cv2.COLOR_RGB2BGR))
```
fcakyon, line 383: We should not remove this line because it is necessary in `export_visual()` method of the `PredictionResult` class: https://github.com/obss/sahi/blob/2824b98c344e052289343cf1a4af23c477325454/sahi/prediction.py#L167

If you dont want output, you can call `visualize_object_predictions` with visualize_object_predictions(output_dir=None)

sahi/predict.py
```diff
@@ -579,9 +687,30 @@ def predict(
...
+            )
+            get_image_from_dict = image_and_elapsed_time["image"]
+
         time_end = time.time() - time_start
         durations_in_seconds["export_files"] = time_end
 
+        if export_video:
+            out.write(get_image_from_dict)
+        else:
+            Path(output_dir).mkdir(parents=True, exist_ok=True)
```
fcakyon, line 710: `output_dir` is not defined when `export_visual=False`. Thus, this line fails when `export_visual=False`.

sahi/predict.py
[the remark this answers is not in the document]
devrimcavusoglu: It's not needed actually (and also a bit confusing) as `fast_forwarding` doesn't only refer to forward skips, but also to backward, so the name should be `fast_forwarding` even if it allows backward skip.

sahi/predict.py
madenburak: Ok. I will edit this.

README.md
madenburak: When I think as Turkish, fast-forward is not enough. If fast-forward is enough in English, we can say fast-forward. I'm not very good at English.

sahi/predict.py
madenburak: Actually, the method that you say is more useful. I will edit the code with this way when I have time.

README.md
```diff
@@ -154,6 +154,33 @@ Find detailed info on COCO utilities (yolov5 conversion, slicing, subsampling, f
...
 
+### Input, Export Video Support 
+
+
+A video can be given as input and the results can be exported as video. You can also see the SAHI predictions during inference on the interactive window where fast-forwarding or backwarding is also possible.
+
+One more thing added prediction time to terminal.
+
+For input, export video:
+` sahi predict --model_path yolov5s.pt --model_config_path coco.yaml --model_type yolov5 --export_video --video_path home/burak/video.mp4`
```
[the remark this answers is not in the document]
madenburak: I will remove `video_path`.  `source` is to be used for both the image and video.

sahi/predict.py
madenburak: I'll do it in a few days.

sahi/utils/cv.py
madenburak: If I don't remove this line, each video frame saving as image in `runs/exp/visual` during `export_visual` and `export_video`. But, if I want to inference on image with `export_visual`, I save result of prediction image in [sahi/sahi/predict.py (line:709)](https://github.com/madenburak/sahi/blob/2824b98c344e052289343cf1a4af23c477325454/sahi/predict.py#L709).

sahi/predict.py
madenburak: I didn't try with `export_visual=False`. I will review this situation.

README.md
devrimcavusoglu: Yes, fast-forward is sufficient in English to cover both meaning.

madenburak:
I did the necessary edits. Can you review it? @fcakyon @devrimcavusoglu

sahi/utils/cv.py
```diff
@@ -210,6 +210,78 @@ def apply_color_mask(image: np.ndarray, color: tuple):
...
+                    print(
+                        "\n===========================Closing==========================="
+                    )  # Exit the prediction, Key = Esc
+                    exit()
+
+                ret, frame = video_capture.read()
+                if not ret:
+                    print("\n===========================Video Ended===========================")
+                    break
+                yield Image.fromarray(frame)
```
devrimcavusoglu, line 277: Thanks for the refactor. My revision here is again, to refactor this inner functions/iterators. I suggest you to create the function `read_video_frame` (independent of this scope) another util function, which takes `video_capture` and `fast_forwarding` as parameters. For example it'd be sth like this,

```python
def read_video_frame(video_capture, fast_forwarding):
    if fast_forwarding:
        ...
        yield ...
    else:
        ...
        yield ...
```

sahi/utils/cv.py
```diff
@@ -210,6 +210,78 @@ def apply_color_mask(image: np.ndarray, color: tuple):
...
+
+    if export_visual:
```
devrimcavusoglu, line 279: Let's delete the if blocks here and only return the following (see also the comment above for `out`) that way we unify the return type which makes the code less prone to mistakes.

```python
return read_video_frame(video_capture), out, image_base
```

sahi/utils/cv.py
```diff
@@ -210,6 +210,78 @@ def apply_color_mask(image: np.ndarray, color: tuple):
...
 
 
+def get_video_reader(
+    source: str, save_dir: str, export_visual: bool = False, view_image: bool = False, fast_forwarding: bool = False
+):
+    # get video name with extension
+    image_base = os.path.basename(source)
+    # get video from video path
+    video_capture = cv2.VideoCapture(source)
+    if export_visual:
```
devrimcavusoglu, line 220: Let's add else for this `if` and set `out` to None (for return).

devrimcavusoglu (changes requested):

sahi/utils/cv.py
fcakyon: Lines between 381-383 should not be deleted in `sahi/utils/cv.py`. It is necessary for other functions and classes.

Instead we should add a check in `predict` function so that `output_dir=None` is set for `visualize_object_predictions` function when `export_video=True` and `export_visual=True`

sahi/predict.py
```diff
@@ -460,14 +473,38 @@ def predict(
...
+
+        read_video_frame, image_base = get_video_reader(source, save_dir, export_visual, view_image, fast_forwarding)
+        item = read_video_frame
+
+    elif input_video and export_visual:
+
+        read_video_frame, out, image_base = get_video_reader(
+            source, save_dir, export_visual, view_image, fast_forwarding
+        )
+        item = read_video_frame
```
fcakyon, line 488: this function should be called as `read_video_frame()`, i believe

sahi/predict.py
```diff
@@ -431,13 +443,14 @@ def predict(
...
         image_path_list = [source]
 
-    # init export directories
-    save_dir = Path(increment_path(Path(project) / name, exist_ok=False))  # increment run
-    crop_dir = save_dir / "crops"
-    visual_dir = save_dir / "visuals"
-    visual_with_gt_dir = save_dir / "visuals_with_gt"
-    pickle_dir = save_dir / "pickles"
-    save_dir.mkdir(parents=True, exist_ok=True)  # make dir
+    if export_visual:
```
fcakyon, line 446: we have to remove this if, they dont depend on `export_visual` param.

save_dir: can be used for coco result json export
crop_dir: is used when export_crop=True`
pickle_dir: is used when `export_pickle=True`

sahi/predict.py
```diff
@@ -315,6 +317,10 @@ def predict(
     postprocess_match_threshold: float = 0.5,
     postprocess_class_agnostic: bool = False,
     export_visual: bool = False,
+    input_video: bool = False,
```
fcakyon, line 320: we should remove this parameter and check if source is video by looking its suffix (mp4, avi, mpeg etc)

sahi/predict.py
```diff
@@ -315,6 +317,10 @@ def predict(
...
+    view_image: bool = False,
+    fast_forwarding: bool = False,
+    save_dir: str = None,
```
fcakyon, line 323: this parameter is not necessary, we can remove it, save_dir is generated as `{project}/{name}` in line 435

madenburak: That's true. I removed it.

sahi/predict.py
madenburak: Good idea for video control.

fcakyon: you can get the suffix as:
```python
from pathlib import Path
suffix = Path(file_path).suffix
```
source: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix

sahi/predict.py
madenburak: We make `read_video_frame()` function that return of `get_video_reader` assign to `read_video_frame` variable. Therefore, we don't use `read_video_frame()`.

sahi/predict.py
madenburak: I have used this if block because when every run sahi, it was creating null output file in `runs/predict.`  We can remove it, if null output file is unimportant. Or we will add  `if export_visual:` on top of `save_dir.mkdir(parents=True, exist_ok=True)  # make dir`

sahi/utils/cv.py
madenburak: I correct it.

sahi/utils/cv.py
madenburak: If we don't use two option of return, every run sahi, it was creating null output file in `runs/predict`. Because of `out` variable.

madenburak: I changed it but I would have to use if block for return.

devrimcavusoglu: why tho ? Isn't this controlled with `export_visual`, so my thought was like this

```python
if export_visual:
    # Regular stuff here
    # get video properties and create VideoWriter object
    fps = video_capture.get(cv2.CAP_PROP_FPS)
    # ...
else:
    out = None

return read_video_frame(video_capture), out, image_base
```

If you think this would behave to create empty output file, can you elaborate why ?

sahi/utils/cv.py
devrimcavusoglu: Here, I was thinking that, why would not we make the fast-forwarding the default and only behavior, does it impose any harm to the users ? wdyt about making it the only way @madenburak @fcakyon ? Fast-forwarding mode already covers the usage of not using the FF mode, or is there any ?

madenburak: You're right. I thinked without FF program is Faster than with FF. But When I reviewed, this situation is unreasonable. FF must be default choice. Also, If num of slice is increase, view image is slowing. For this, I thinking, we can process some frames(users choice) instead of all frames. For examle, one of every 3(user’s choice) frames. If user don’t want to this, can choice 1.

sahi/utils/cv.py
madenburak: I am interested in this situation.

sahi/predict.py
fcakyon: @madenburak this line should be updated as:
```python
if export_visual or export_pickle or export_crop or dataset_json_path is not None:
```

fcakyon (changes requested):

fcakyon:
@madenburak thanks a lot for all your effort, we are close to finish!

madenburak:
> @madenburak thanks a lot for all your effort, we are close to finish!

It has been my pleasure.

madenburak:
Can you review new changes, please?. @fcakyon @devrimcavusoglu

sahi/predict.py
madenburak: > @madenburak this line should be updated as:
> 
> ```python
> if export_visual or export_pickle or export_crop or dataset_json_path is not None:
> ```

I am correcting, now.

devrimcavusoglu (approved):

sahi/utils/cv.py
fcakyon: @madenburak these changes should be reverted

fcakyon (changes requested):

madenburak:
I added `if output_dir:` in cv.py, but i don't used it as `if output_dir:`. Because we have more than one option, now.

madenburak:
Are there any parts that need to be changed?

fcakyon:
@madenburak i will perform some local testing and return back to you :+1:

fcakyon:
@madenburak can you add me as contributor to your fork? I will push some updates

madenburak:
I added you to my fork.

fcakyon (approved):

fcakyon:
@madenburak thanks a lot for this amazing contribution!

PR #445, opened 2022-04-14
reduce ram usage by adding buffer based merging

weypro:
Add merge_buffer_length


fcakyon:
@weypro can you please reformat your code as described in: https://github.com/obss/sahi#contributing

devrimcavusoglu:
Can we also refactor the image slicing part, np slicing is no worse than creating new PIL.Image objects and presumably better in some scenarios, so I also suggest that we do that refactor in this PR as well (as the discussion originated from this point, it should be a part of this PR). Wdyt ? @weypro @fcakyon

fcakyon:
> Can we also refactor the image slicing part, np slicing is always better than creating new PIL.Image objects even though there is slight improvement in some cases, so I also suggest that we do that refactor in this PR as well (as the discussion originated from this point, it should be a part of this PR). Wdyt ? @weypro @fcakyon

Yes we should definitely open numpy slice improvement as a new PR since it does not have any effect on detection accuracy it is an easier PR :+1:

fcakyon:
Hey @weypro! Are you still working on this draft?

weypro:
> Hey @weypro! Are you still working on this draft?

Sorry, I have to resume my job. Could you help me finish the rest of the work?

weypro:
I try to explain the problem with this method clearly. Before explaining my thought, I need to mention that I just simply regard NMM and NMS as the same thing, and I will take NMS as the example.

The description of the NMS algorithm from the link https://towardsdatascience.com/non-maximum-suppression-nms-93ce178e177c:
```
1. Select the proposal with highest confidence score, remove it from B and add it to the final proposal list D. (Initially D is empty).
2. Now compare this proposal with all the proposals — calculate the IOU (Intersection over Union) of this proposal with every other proposal. If the IOU is greater than the threshold N, remove that proposal from B.
3. Again take the proposal with the highest confidence from the remaining proposals in B and remove it from B and add it to D.
4. Once again calculate the IOU of this proposal with all the proposals in B and eliminate the boxes which have high IOU than threshold.
This process is repeated until there are no more proposals left in B.
```
Let's notice the phrase `eliminate the boxes which have high IOU than threshold` in the third step. In other words, the boxes  will definitely be eliminated under the following conditions:
1. The eliminated box overlaps with the `highest` proposal. (Actually, it's `the local highest` for every slice)
2. The IOU is higher than the threshold.  

The second condition depends on the first one, and all the problems are caused by `the local highest`.
What we concern is whether the eliminated proposal should not be eliminated and whether the proposal which needs to be eliminated does not be eliminated.
1. For the first question, the phenomenon is the IOU of the accidentally eliminated proposal and `the local highest` is higher than the threshold, and `the local highest` is also eliminated in later iterations. But the IOU of the first eliminated proposal with other proposals is lower than the threshold, which means it needn't be eliminated.
For example, we have three horizontally arranged proposals with scores of 0.5, 0.6, and 0.7. The IOUs are 0.5 and the threshold is 0.4. The first two proposals are in the first slice and the rest is in the second slice.
When we apply NMS to the whole list of proposals, we can only eliminate the second proposal and we finally get two proposals as the result.
When we try to apply NMS separately, the first proposal is eliminated in the first iteration, and the second proposal is also eliminated in the second iteration. The number of iterations equals the number of slices.
The root cause is the accidentally eliminated proposals are associated with proposals outside the scope. To solve it, a simple method is that we can limit the proposal in the central area to preprocess in advance, while it doesn't really solve the problem of insufficient memory.
2. For the second question, it should not exist in theory, because all of the new proposals will be included in the comparison scope in every iteration.

The conclusion is that the method can just alleviate the problem of insufficient memory, and it may be better to try to save the cache to the hard disk as the final solution.

devrimcavusoglu:
>    For example, we have three horizontally arranged proposals with scores of 0.5, 0.6, and 0.7. The IOUs are 0.5 and the threshold is 0.4. The first two proposals are in the first slice and the rest is in the second slice.
>    When we apply NMS to the whole list of proposals, we can only eliminate the second proposal and we finally get two proposals as the result.
>    When we try to apply NMS separately, the first proposal is eliminated in the first iteration, and the second proposal is also eliminated in the second iteration. The number of iterations equals the number of slices.

@weypro sorry for jumping in late, I think I can take it from here. For the problem above, I have something in mind though I am not 100% sure if it's feasible or not, and probably it will introduce a little bit more memory overhead. It's not a direct solution, but kinda work-around. If we populate proposals from non-adjacent slices in buffer, then'd be **less worried** about accidental eliminations. Note that, I said "less worried" as it doesn't introduce the mutual exclusion between overlapping proposals, still there is a little room for accidental eliminations (but far less compared to current implementation in practice). This is mainly due to overlapping behavior of slices. In theory there is a way to introduce a full disjoint slices in buffer by taking slice locations into account.

weypro:
> > For example, we have three horizontally arranged proposals with scores of 0.5, 0.6, and 0.7. The IOUs are 0.5 and the threshold is 0.4. The first two proposals are in the first slice and the rest is in the second slice.
> > When we apply NMS to the whole list of proposals, we can only eliminate the second proposal and we finally get two proposals as the result.
> > When we try to apply NMS separately, the first proposal is eliminated in the first iteration, and the second proposal is also eliminated in the second iteration. The number of iterations equals the number of slices.
> 
> @weypro sorry for jumping in late, I think I can take it from here. For the problem above, I have something in mind though I am not 100% sure if it's feasible or not, and probably it will introduce a little bit more memory overhead. It's not a direct solution, but kinda work-around. If we populate proposals from non-adjacent slices in buffer, then'd be **less worried** about accidental eliminations. Note that, I said "less worried" as it doesn't introduce the mutual exclusion between overlapping proposals, still there is a little room for accidental eliminations (but far less compared to current implementation in practice). This is mainly due to overlapping behavior of slices. In theory there is a way to introduce a full disjoint slices in buffer by taking slice locations into account.

Yes. But I think it's unnecessary to implement this because we have already gotten a better solution to avoid the problem that is saving cache to the hard disk. It seems like a waste of time if we continue to work on the problem.

devrimcavusoglu:
@weypro @fcakyon results on `merge_buffer_length` for various configs. The dataset used is this [sample_dataset](https://github.com/obss/sahi/releases/download/0.8.19/sample_dataset.zip). Presumably, for this dataset `None` and `100` have identical eval results. If we are okay with the trade-off, we can proceed with this implementation. 

```python
 merge_buffer_length = None
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.362
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.576
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.368
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.551
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.673
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.625
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.301
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.452
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.463
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.371
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.502
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.650

 merge_buffer_length = 1
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.135
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.228
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.145
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.208
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.265
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.500
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.089
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.162
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.108
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.176
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400

 merge_buffer_length = 3
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.140
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.233
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.150
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.208
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.273
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.500
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.089
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.169
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.108
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.184
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400

 merge_buffer_length = 5
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.142
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.235
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.152
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.208
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.277
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.500
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.089
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.172
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.108
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.188
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400

 merge_buffer_length = 10
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.146
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.241
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.157
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.208
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.287
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.500
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.089
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.179
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.108
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.195
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400

 merge_buffer_length = 25
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.170
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.276
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.186
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.265
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.322
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.500
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.123
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.205
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.148
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.226
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400

 merge_buffer_length = 50
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.251
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.407
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.249
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.411
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.459
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.500
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.209
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.299
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.266
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.346
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.400

 merge_buffer_length = 100
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=500 ] = 0.362
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=500 ] = 0.576
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=500 ] = 0.368
 Average Precision  (AP) @[ IoU=0.50      | area= small | maxDets=500 ] = 0.551
 Average Precision  (AP) @[ IoU=0.50      | area=medium | maxDets=500 ] = 0.673
 Average Precision  (AP) @[ IoU=0.50      | area= large | maxDets=500 ] = 0.625
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.301
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.452
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.463
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=500 ] = 0.371
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=500 ] = 0.502
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=500 ] = 0.650
```

fcakyon:
Thanks for the effort @devrimcavusoglu! We can merge it considering None and 100 have similar results! Should we add any info to the docstring related wirth the new parameter?

devrimcavusoglu:
> Thanks for the effort @devrimcavusoglu! We can merge it considering None and 100 have similar results! Should we add any info to the docstring related wirth the new parameter?

Yes we definitely need to add an info about the trade-off to docstring. I'll add in few hours, and we can merge.

sahi/predict.py
```diff
@@ -165,6 +166,9 @@ def get_sliced_prediction(
             0: no print
             1: print number of slices (default)
             2: print number of slices and slice/prediction durations
+        merge_buffer_length: int
+            The length of buffer for slices to be used during sliced prediction, which is suitable for low memory.
+            It may affect the AP if it is specified.
```
devrimcavusoglu, line 171:
```suggestion
            It may affect the AP if it is specified. The higher the amount, the closer results to the non-buffered
            scenario. See [the discussion](https://github.com/obss/sahi/pull/445).
```

devrimcavusoglu (changes requested):

sahi/predict.py
devrimcavusoglu: @weypro @fcakyon can you check the docstring and commit if it's okay ?

devrimcavusoglu (approved):

fcakyon (approved):

fcakyon:
@devrimcavusoglu @weypro thanks for the effort!

PR #462, opened 2022-05-11
add support to instantiate a `DetectionModel` from layer

mecevit:
This PR adds support to instantiate a DetectionModel out of a [YOLOv5](https://github.com/ultralytics/yolov5) model trained on [Layer](https://layer.ai). It includes:
- A new `from_layer()` method in `AutoDetectionModel` to load models from Layer.
- A new `set_model()` method in `DetectionModel` to create a `DetectionModel` out of an already loaded model.
- Tests for `from_layer()` and `set_model()`.

### Example
You can pass a path of a YOLOv5 model trained on Layer to instantiate a `DetectionModel`.  Here we fetch the [`yolo5vs` pretrained model](https://app.layer.ai/layer/yolov5/models/yolov5s) from [this Layer project](https://app.layer.ai/layer/yolov5)

```python
from sahi.model import DetectionModel
detection_model = DetectionModel.from_layer("layer/yolov5/models/yolov5s")
```

### Tests
To run the tests related to `from_layer()` functionality you can:
```
pytest tests/test_layer.py
```

### Screenshot

image


setup.py
```diff
@@ -36,7 +36,7 @@ def get_version():
     python_requires=">=3.7",
     install_requires=get_requirements(),
     extras_require={
-        "tests": ["pytest", "mmdet", "norfair"],
+        "tests": ["pytest", "mmdet", "norfair", "layer"],
```
fcakyon, line 39: We should also update the [ci workflow file](https://github.com/mecevit/sahi/blob/main/.github/workflows/ci.yml) by this addition: 

```yaml
      - name: Install YOLOv5(6.1.2) and Norfair(0.4.3)
        run: >
          pip install yolov5==6.1.2 norfair==0.4.3
```

mecevit: Done: https://github.com/obss/sahi/pull/462/commits/cf9ddbe1ebc3b4669307aaa555ee28c9ab95eefb

fcakyon:
We are trying to finalize https://github.com/obss/sahi/pull/368 before merging PR 👍

fcakyon:
We now have a factory class `AutoDetectionModel` as: https://github.com/obss/sahi/blob/02949de39815841476f5053452ac76417685c502/sahi/auto_model.py#L13

It would be better to move this `.from_layer()` method from base class to the new `AutoDetectionModel` class as:

```python
    @staticmethod
    def from_layer(
        model_path: str,
        no_cache: bool = False,
        device: Optional[str] = None,
        mask_threshold: float = 0.5,
        confidence_threshold: float = 0.3,
        category_mapping: Optional[Dict] = None,
        category_remapping: Optional[Dict] = None,
        image_size: int = None,
    ):
        """
        Loads a DetectionModel from Layer. You can pass additional parameters in the name to retrieve a specific version
        of the model with format: ``model_path:major_version.minor_version``
        By default, this function caches models locally when possible.
        Args:
            model_path: str
                Path of the Layer model (ex. '/sahi/yolo/models/yolov5')
            no_cache: bool
                If True, force model fetch from the remote location.
            device: str
                Device, "cpu" or "cuda:0"
            mask_threshold: float
                Value to threshold mask pixels, should be between 0 and 1
            confidence_threshold: float
                All predictions with score < confidence_threshold will be discarded
            category_mapping: dict: str to str
                Mapping from category id (str) to category name (str) e.g. {"1": "pedestrian"}
            category_remapping: dict: str to int
                Remap category ids based on category names, after performing inference e.g. {"car": 3}
            image_size: int
                Inference input size.
        Returns:
            Returns an instance of a DetectionModel
        Raises:
            ImportError: If Layer is not installed in your environment
            ValueError: If model path does not match expected pattern: organization_name/project_name/models/model_name
        """
        try:
            import layer
        except ImportError:
            raise ImportError('Please run "pip install layer -U" ' "to load models from Layer.")

        layer_model = layer.get_model(name=model_path, no_cache=no_cache).get_train()
        if layer_model.__class__.__module__ in ["yolov5.models.common", "models.common"]:
            model_type = "yolov5"
        else:
            raise Exception(f"Unsupported model: {type(layer_model)}. Only YOLOv5 models are supported.")

        DetectionModel = MODEL_TYPE_TO_MODEL_CLASS_NAME[model_type]
        return DetectionModel(
            model=layer_model,
            device=device,
            mask_threshold=mask_threshold,
            confidence_threshold=confidence_threshold,
            category_mapping=category_mapping,
            category_remapping=category_remapping,
            image_size=image_size,
        )

```

Then it can be called as:

```python

yolov5_detection_model = AutoDetectionModel.from_layer('/sahi/yolo/models/yolov5')
```

fcakyon (approved):

fcakyon:
@mecevit thanks for this great pr! its now merged.

i will work on a notebook showcasing the use of `.from_layer()` when i find the time. once its ready, we can release a new `sahi` version that includes the `layer` support :+1:

PR #465, opened 2022-05-21
Added support for exporting predictions in COCO format

PushpakBhoge:
# Added support for exporting predictions in COCO format 

## changelog: -
* COCO prediction object of class `CocoPrediction` can be added into COCO image object of class `sahi.utils.cocoCocoImage` using method `<CocoImage_Instance>.add_prediction`
* `__repr__` method updated to reflect predictions in class `sahi.utils..coco.CocoImage`
* Additional property `prediction_array` added in class `sahi.utils.coco.Coco` to export COCO predictions in json-serializable format. Logic for the property is defined in new function `sahi.utils.coco.create_coco_prediction_array` 
* added the test cases for the same in `test_cocoutils.py`
* updated file `coco.md` with necessary documentation

#### Basic flow is like this
```python
from sahi.utils.coco import Coco, CocoImage, CocoAnnotation, CocoPrediction
from sahi.utils.file import save_json

coco_obj = Coco()

# add n images to coco_obj
for _ in range(n):
    image = CocoImage(**kwargs)
    
    # add n annotations to the image
    for _ in ange(n):
        image.add_annotation(CocoAnnotation(**kwargs))
    
    # add n predictions to the image
    for _ in range(n)
        image.add_prediction(CocoPrediction(**kwargs))
    
    # add image to coco object
    coco_obj.add_image(image)

# export ground truth annotations
coco_gt = coco_obj.json
save_json(coco_gt , "ground_truth.json")

# export preadictions 
coco_predictions = coco_obj.prediction_array
save_json(coco_predictions, "predictions.json"
```

## Why this is useful?
The user can utilize the exported files to calculate standard coco metrics using offical COCO api `pycocotool` with relative ease.
This is example of using that
```python
from pycocotools.cocoeval import COCOeval
from pycocotools.coco import COCO

# load json files
coco_ground_truth = COCO(annotation_file="ground_truth.json")
coco_predictions = coco_ground_truth.loadRes("predictions.json")

coco_evaluator = COCOeval(coco_ground_truth, coco_predictions, "bbox")
coco_evaluator.evaluate()
coco_evaluator.accumulate()
coco_evaluator.summarize()
```
This can be further expanded to define class which can be conveniently used to evaluate the prediction, something like this
```python
class DatasetEvalutor:
    # user can inherit and overwrite the given load_model, get_annotation and get_prediction methods
    def __init__(self, **kwargs):
        self.images = []  # list of image meta data
        self.model = None # model place holder
        self.dataset = Coco() # coco object

    def load_model(self,**kwargs):
        # logic for loading model goes here
        # set self.model to model

    def get_annotation(self, **kwargs):
        # logic for annotation goes here
        # result must return a sahi.utils.coco.CocoAnnotation instance

    def get_prediction(self, **kwargs):
        # logic for prediction goes here
        # result must return a sahi.utils.coco.CocoPrediction instance

    def run_inference(self, **kwargs):
        # loop over images create CocoImage and add annotations and prediction to it
        # add those CocoImage instance into self.dataset

    def evaluate(self, **kwargs):
        # pycocotool code to generate metrics
```

### DIFF
--- a/docs/coco.md
+++ b/docs/coco.md
@@ -62,7 +62,7 @@ coco_image.add_prediction(
 coco_image.add_prediction(
   CocoPrediction(
     score=0.653424,
-    segmentation=[x_min, y_min, width, height],
+    bbox=[x_min, y_min, width, height],
     category_id=1,
     category_name='vehicle'
   )
--- a/docs/coco.md
+++ b/docs/coco.md
@@ -48,6 +48,26 @@ coco_image.add_annotation(
   )
 )
 ```
+- add predictions to coco image:
+
+```python
+coco_image.add_prediction(
+  CocoPrediction(
+    score=0.864434,
+    bbox=[x_min, y_min, width, height],
+    category_id=0,
+    category_name='human'
+  )
+)
+coco_image.add_prediction(
+  CocoPrediction(
+    score=0.653424,
+    segmentation=[x_min, y_min, width, height],
+    category_id=1,
+    category_name='vehicle'
+  )
+)
+```
 
 - add coco image to Coco object:
 
@@ -68,6 +88,32 @@ from sahi.utils.file import save_json
 
 save_json(coco_json, "coco_dataset.json")
 ```
+- you can also export prediction array in coco prediction format and save it as json :
+
+```python
+from sahi.utils.file import save_json
+
+predictions_array = coco.prediction_array
+save_json = save_json(predictions_array, "coco_predictions.json")
+```
+- this prediction array can be used to get standard coco metrics for the predictions using official pycocotool api :
+
+```python
+# note:- pycocotools need to be installed seperately 
+from pycocotools.cocoeval import COCOeval
+from pycocotools.coco import COCO
+
+coco_ground_truth = COCO(annotation_file="coco_dataset.json")
+coco_predictions = coco_ground_truth.loadRes("coco_predictions.json")
+
+coco_evaluator = COCOeval(coco_ground_truth, coco_predictions, "bbox")
+coco_evaluator.evaluate()
+coco_evaluator.accumulate()
+coco_evaluator.summarize()
+```
+
+
+
 </details>
 
 <details closed>
--- a/sahi/utils/coco.py
+++ b/sahi/utils/coco.py
@@ -620,8 +620,8 @@ class CocoImage:
     file_name: {self.file_name},
     height: {self.height},
     width: {self.width},
-    annotations: List[CocoAnnotation]>,
-    predictions: List[CocoPrediction]"""
+    annotations: List[CocoAnnotation],
+    predictions: List[CocoPrediction]>"""
 
 
 class CocoVidImage(CocoImage):
--- a/sahi/utils/coco.py
+++ b/sahi/utils/coco.py
@@ -599,7 +599,6 @@ class CocoImage:
         """
         Adds prediction to this CocoImage instance
 
-        annotation : CocoAnnotation
         prediction : CocoPrediction
         """
 
--- a/sahi/utils/coco.py
+++ b/sahi/utils/coco.py
@@ -583,6 +583,7 @@ class CocoImage:
         self.height = int(height)
         self.width = int(width)
         self.annotations = []  # list of CocoAnnotation that belong to this image
+        self.predictions = []  # list of CocoPrediction that belong to this image
 
     def add_annotation(self, annotation):
         """
@@ -593,6 +594,17 @@ class CocoImage:
 
         assert isinstance(annotation, CocoAnnotation), "annotation must be a CocoAnnotation instance"
         self.annotations.append(annotation)
+
+    def add_prediction(self, prediction):
+        """
+        Adds prediction to this CocoImage instance
+
+        annotation : CocoAnnotation
+        prediction : CocoPrediction
+        """
+
+        assert isinstance(prediction, CocoPrediction), "prediction must be a CocoPrediction instance"
+        self.predictions.append(prediction)
 
     @property
     def json(self):
@@ -609,7 +621,8 @@ class CocoImage:
     file_name: {self.file_name},
     height: {self.height},
     width: {self.width},
-    annotations: List[CocoAnnotation]>"""
+    annotations: List[CocoAnnotation]>,
+    predictions: List[CocoPrediction]"""
 
 
 class CocoVidImage(CocoImage):
@@ -1067,6 +1080,13 @@ class Coco:
             images=self.images,
             categories=self.json_categories,
             ignore_negative_samples=self.ignore_negative_samples,
+        )
+
+    @property
+    def prediction_array(self):
+        return create_coco_prediction_array(
+            images=self.images,
+            ignore_negative_samples=self.ignore_negative_samples
         )
 
     @property
@@ -1885,6 +1905,58 @@ def create_coco_dict(images, categories, ignore_negative_samples=False):
     }
     # return coco dict
     return coco_dict
+
+
+def create_coco_prediction_array(images, ignore_negative_samples=False):
+    """
+    Creates COCO prediction array which is list of predictions
+
+    Arguments
+    ---------
+        images : List of CocoImage containing a list of CocoAnnotation
+        ignore_negative_samples : Bool
+            If True, images without predictions are ignored
+    Returns
+    -------
+        coco_prediction_array : List
+            COCO predictions array
+    """
+    predictions_array = []
+
+    num_images = len(images)
+    image_id = 1
+    prediction_id = 1
+    for image_ind in range(num_images):
+        # get coco image and its coco predictions
+        coco_image = images[image_ind]
+        coco_predictions = coco_image.predictions
+        # get num annotations
+        num_predictions = len(coco_predictions)
+        # if ignore_negative_samples is True and no annotations, skip image
+        if ignore_negative_samples and num_predictions == 0:
+            continue
+        else:
+            # create coco prediction object
+            for coco_prediction in coco_predictions:
+                # create coco prediction object
+                out_prediction = {
+                    "id": prediction_id,
+                    "image_id": image_id,
+                    "bbox": coco_prediction.bbox,
+                    "score": coco_prediction.score,
+                    "category_id": coco_prediction.category_id,
+                    "segmentation": coco_prediction.segmentation,
+                    "iscrowd": coco_prediction.iscrowd,
+                    "area": coco_prediction.area,
+                }
+                predictions_array.append(out_prediction)
+                prediction_id = prediction_id + 1
+
+            # increment annotation id
+            image_id = image_id + 1
+
+    # return predictions array
+    return predictions_array
 
 
 def add_bbox_and_area_to_coco(
--- a/sahi/utils/coco.py
+++ b/sahi/utils/coco.py
@@ -798,6 +798,7 @@ class Coco:
         remapping_dict=None,
         ignore_negative_samples=False,
         clip_bboxes_to_img_dims=False,
+        image_id_setting="auto",
     ):
         """
         Creates Coco object.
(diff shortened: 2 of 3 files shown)


fcakyon:
@PushpakBhoge thanks for this great contribution!

Can you please reformat your code so that it passes linting tests?

- Install required development packages:

```bash
pip install -e ."[dev]"
```

- Reformat with black and isort:

```bash
black . --config pyproject.toml
isort .
```

PushpakBhoge:
@fcakyon done followed the commands reformatted

sahi/utils/coco.py
```diff
@@ -1887,6 +1903,58 @@ def create_coco_dict(images, categories, ignore_negative_samples=False):
...
+            If True, images without predictions are ignored
+    Returns
+    -------
+        coco_prediction_array : List
+            COCO predictions array
+    """
+    predictions_array = []
+
+    num_images = len(images)
+    image_id = 1
```
fcakyon, line 1923: image ids should not be hard coded but fetched from the image objects as `image.id`. image ids dont have to be ordered.

fcakyon (changes requested):

sahi/utils/coco.py
fcakyon: @PushpakBhoge  to fix this you can iterate over images instead of range()

PushpakBhoge: @fcakyon initially I thought of doing it in the similar way, but turns out `image_id` is not stored in instance of `CocoImage`. the `image_id` has been always `None` in those instances, I have verified this when I was working on some project. If we do it like this then image_id in json will be `null` that is invalid and will not work with `pycocotool`

I looked at function `sahi.utils.coco.create_coco_dict` which is associated with property `sahi.utils.Coco.json`. This function contain logic to generate Ground Truth coco json. This function work in the similar manner it hard code image_id and then increment it inside the loop. I am not aware why it is like this but I guess it might be required during coco slicing operation.

Hence I followed the similar structure and wrote logic in `sahi.utils.coco.create_coco_prediction_array`, trying not to modify existing flow. What would be your suggestion in this case ?

fcakyon: @PushpakBhoge by default, id property is set as None, your are right on that.

However, when creating a CocoImage instance from a coco image json, id property is automatically set: https://github.com/obss/sahi/blob/79e28fecf48041b6c62fbca8e85b52ed71fa0375/sahi/utils/coco.py#L552

As a result, when instantiating a Coco object by dataset json path with this method: https://github.com/obss/sahi/blob/79e28fecf48041b6c62fbca8e85b52ed71fa0375/sahi/utils/coco.py#L964

all image CocoImage instances include valid `id`s

So you have to instantiate Coco instance as:

```python
from sahi.utils.coco import Coco
coco_obj = Coco.from_coco_dict_or_path("ground_truth.json")
```

Then all instances in `coco_obj.images` will include `id` property.

fcakyon: `sahi.utils.coco.create_coco_dict` is designed to export a custom created coco dataset so it assigns the ids from scratch.

It is not the case when working with predictions (all images ids already include id)

PushpakBhoge: ok got what you are saying

we have "ground_truth.json" in coco format we load it in `coco_obj`, populate the predictions and then export predictions keeping image_ids in sync. This will work seamlessly if we added image ids from `image.id` 

Consider this we don't have annotations in coco format i.e. 'ground_truth.json', but we have annotations in some other format yolo, pascal or any else. Here to get coco metrics we need both predictions and ground truths in coco format. So in this case we will need to export both `json` and `predicion_array` properties right ?

Now, `json` property will assign the ids from scratch and `prediction_array` will look for ids in Image id in `CocoImage` which will be `None`

let's say we assigned ids at the time of creating `CocoImage` using `from_coco_image_dict`  method. In this case `CocoImage` object will have ids. `prediction_array` property will pick up those ids but `json` property will still reassign from scratch, won't that affect synch of image ids between the two json?

So, I think either we fetch id from `CocoImage` for both properties or we assign the ids from scratch for both properties. otherwise there might occur issues where the ids are not in sync.

for the first option we will have to make `<CocoImage_instance>.id` attribute as `NOT NULL`  that might have larger impact on other aspects of the framework. second option is simpler and have no impact on other functionality and both `json` and `prediction_array` can be thought of as temporary files to get coco-metrics. If anyone want to traceback to original ids a simple mapping using image name will be easy to create.

This was my thinking process to treat those two properties as temporary files, but if you think that we need to add ids from `CocoImage` will be happy to do that. let me know what are your thoughts.

P.S. - sorry for a very looooooooooooooong comment :bowtie:

fcakyon: Thanks a lot @PushpakBhoge for your detailed explanations!

> Consider this we don't have annotations in coco format i.e. 'ground_truth.json', but we have annotations in some other format yolo, pascal or any else. Here to get coco metrics we need both predictions and ground truths in coco format. So in this case we will need to export both `json` and `predicion_array` properties right ?
> 
> Now, `json` property will assign the ids from scratch and `prediction_array` will look for ids in Image id in `CocoImage` which will be `None`

This is an awesome point!

Then we need to implement `create_coco_prediction_array` and `create_coco_dict` in a similar fashion that:
1. they iterate over CocoImage instances instead of integer range
2. when `image.id is None` then generate image_id's from a counter
3. when `image.id is not None` use `image.id` as the image_id

So `create_coco_dict` also requires an update (i can update it after this PR).

PushpakBhoge: okie great that sounds like a good idea

There is one thing that might go wrong, what if used added some images which have ids and some images which do not have ids ? This might create duplicate ids, incorrect id mapping etc.

This is off course invalid input scenario, but do we need to came up with a way to handle this. Maybe restricting the user in some way or is it fine if we mention the same thing in documentation and trust user to do right thing ?

fcakyon: How about this:

when `create_coco_prediction_array ` or `create_coco_dict ` is called:
1. create a `image_id_set` using the id properties of the image instances (for the instances that contain id)
2. iterate over CocoImage instances instead of integer range
2.1. when image.id is None then generate image_id's from a counter, if generated id is in `image_id_set` increment again till generated number is not in `image_id_set`
2.2. when image.id is not None use image.id as the image_id

This way we can handle even if user forgets to assign id to some of the image instances.

PushpakBhoge: Sorry got busy

There is one case that might come. Let's we are assigning with counter and the counter reaches 100 then user send image with id 6, but that id is already assigned. This will create duplicate image_id which will make the coco invalid

I tried thinking of way how we can handle this but seems like there not much we can do without overengineering, hers's what I propose

1) We will recommend user in docs to add all `CocoImage` with existing image ids in incremental order first, then `CocoImage` with `image.id = None`
2) we will initialize a variable in `<Coco_Instance>.__init__` as `self.counter=1`, each time we receive image id if `image.id >= self.counter` we will set `self.counter = image_id + 1` othereise if `image_id < self.counter` we will throw exception saying `{image.id}  already taken, add CocoImage with predefined image_ids in incremental order first before adding CocoImage without image_ids to avoid this error` or something in line
3) in second stage when user add `CocoImage` with `image.id=None` we will internally increment the `self.counter` and set that in the `CocoImage` instance

so even if user added a duplicate id it won't go unnoticed and help keeing the exported the coco valid, since we are adding the ids during `<Coco_Instance>.add_image` user can quickly correct it then and there. Also techinically if image is a part of coco dataset then its better each image also have assigned id at the time of addition rather than waiting for it to export

granted that for some cases even if image_id is not present in the list it will throw exception
example `self.counter=1` and we added image with id=250 then we basically setting `self.counter=251` and it will basically not allow any image with id <=250. since the concept of primery_key (image_id) is that if assigned_id is 250 then there must images with ids upto 250 in original dataset. So this will serve the purpose better

so basically we will modify the add_image method like this
```python
     def add_image(self, image):
        """
        Adds image to this Coco instance

        Args:
            image: CocoImage
        """

        assert type(image) == CocoImage, "image must be a CocoImage instance"

        if image.id:
            if image.id >= self.counter:
                self.counter = image.id + 1
            else:
                raise Exception("{image.id}  already taken, add CocoImage with predefined image_ids in incremental order first before adding CocoImage without image_ids to avoid this error")
        else:
            image.id = self.counter
            self.counter+=1

        self.images.append(image)
```

fcakyon: @PushpakBhoge you are right, we dont have to overengineer this.

I believe this might be simpler:

Add `image_id_setting: str` property to `Coco` instance init. Its value is `"auto"` by default. It can be set as "manual" by choice.

When `image_id_setting=="auto"`, then Image.id is ignored and image ids are generated automatically at export (as its the case now)

When `image_id_setting=="manual"`, then all image ids have to be manually set and `.add_image()` throws error when id is `None`

What do you think?

PushpakBhoge: @fcakyon  yes that sounds like perfect will make the necessary changes

PushpakBhoge:
@fcakyon  made the changes review once

PushpakBhoge:
fixed the bugs which caused failure in unit testing

fcakyon:
@PushpakBhoge this idea also seems to be super interesting! We can work on such PR after merging this one 👍 

> This can be further expanded to define a class which can be conveniently used to evaluate the prediction, something like this
> 
> ```python
> class DatasetEvalutor:
>     # user can inherit and overwrite the given load_model, get_annotation and get_prediction methods
>     def __init__(self, **kwargs):
>         self.images = []  # list of image meta data
>         self.model = None # model place holder
>         self.dataset = Coco() # coco object
> 
>     def load_model(self,**kwargs):
>         # logic for loading model goes here
>         # set self.model to model
> 
>     def get_annotation(self, **kwargs):
>         # logic for annotation goes here
>         # result must return a sahi.utils.coco.CocoAnnotation instance
> 
>     def get_prediction(self, **kwargs):
>         # logic for prediction goes here
>         # result must return a sahi.utils.coco.CocoPrediction instance
> 
>     def run_inference(self, **kwargs):
>         # loop over images create CocoImage and add annotations and prediction to it
>         # add those CocoImage instance into self.dataset
> 
>     def evaluate(self, **kwargs):
>         # pycocotool code to generate metrics
> ```

docs/coco.md
```diff
@@ -48,6 +48,26 @@ coco_image.add_annotation(
...
+    score=0.864434,
+    bbox=[x_min, y_min, width, height],
+    category_id=0,
+    category_name='human'
+  )
+)
+coco_image.add_prediction(
+  CocoPrediction(
+    score=0.653424,
+    segmentation=[x_min, y_min, width, height],
```
fcakyon, line 65: `segmentation` format is a bit different. Should be in the form of `[polygon1, polygon2, ...]` where polygons are in the form [x1, y1, x2, y2, x3, y3, ...]

It's too complex to show in an example, maybe we can directly delete the segmentation example in docs?

sahi/utils/coco.py
```diff
@@ -1828,36 +1858,47 @@ def create_coco_dict(images, categories, ignore_negative_samples=False):
...
     Returns
     -------
         coco_dict : Dict
             COCO dict with fields "images", "annotations", "categories"
     """
-    out_images = []
-    out_annotations = []
-    out_categories = categories
+    # assertion of parameters
+    assert image_id_setting in ["auto", "manual"]
```
fcakyon, line 1871:
```suggestion
    if image_id_setting not in ["auto", "manual"]:
        raise ValueError(f"'image_id_setting' should be one of ['auto', 'manual']")
```

PushpakBhoge:
Hi @fcakyon  made those changes
and I would love to contribute for the COCOevaluator will start preparing some code 
little busy here so maybe raise another PR with initial draft in 1-2 weeks 

thank you

sahi/utils/coco.py
```diff
@@ -609,7 +620,8 @@ def __repr__(self):
     file_name: {self.file_name},
     height: {self.height},
     width: {self.width},
-    annotations: List[CocoAnnotation]>"""
+    annotations: List[CocoAnnotation]>,
+    predictions: List[CocoPrediction]"""
```
devrimcavusoglu, line 624: Change as

```python
annotations: List[CocoAnnotation],
predictions: List[CocoPrediction]>
```

fcakyon, line 624:
```suggestion
    predictions: List[CocoPrediction]>"""
```

sahi/utils/coco.py
```diff
@@ -609,7 +620,8 @@ def __repr__(self):
     file_name: {self.file_name},
     height: {self.height},
     width: {self.width},
-    annotations: List[CocoAnnotation]>"""
+    annotations: List[CocoAnnotation]>,
```
fcakyon, line 623:
```suggestion
    annotations: List[CocoAnnotation],
```

fcakyon:
we are good to merge I believe, thanks for the effort @PushpakBhoge !

fcakyon (approved):

devrimcavusoglu (approved):

PR #469, opened 2022-05-28
add automodel structure for unified model loading

fcakyon:
New `AutoDetectionModel` class makes it easier to load any `yolov5`,  `mmdet`, `detectron2` model!

```python

# load yolov5 models
detection_model = AutoDetectionModel.from_local(
    model_type='yolov5',
    model_path='yolov5s.pt',
    image_size=640,
)

# load mmdet models
detection_model = AutoDetectionModel.from_local(
    model_type='mmdet',
    model_path='model.pt',
    model_config='config.py',
    image_size=640,
)

# load detectron2 models
detection_model = AutoDetectionModel.from_local(
    model_type='detectron2',
    model_path='model.pt',
    model_config='config.yaml',
    image_size=640,
)
```

### DIFF
--- a/sahi/auto_model.py
+++ b/sahi/auto_model.py
@@ -1,0 +1,72 @@
+from typing import Dict, Optional
+
+from sahi.utils.file import import_model_class
+
+MODEL_TYPE_TO_MODEL_CLASS_NAME = {
+    "mmdet": "MmdetDetectionModel",
+    "yolov5": "Yolov5DetectionModel",
+    "detectron2": "Detectron2DetectionModel",
+    "torchvision": "TorchVisionDetectionModel",
+}
+
+
+class AutoDetectionModel:
+    @staticmethod
+    def from_local(
+        model_type: str,
+        model_path: str,
+        config_path: Optional[str] = None,
+        device: Optional[str] = None,
+        mask_threshold: float = 0.5,
+        confidence_threshold: float = 0.3,
+        category_mapping: Optional[Dict] = None,
+        category_remapping: Optional[Dict] = None,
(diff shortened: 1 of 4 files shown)


sahi/predict.py
```diff
@@ -540,7 +535,7 @@ def predict(
         tqdm.write("Prediction time is: {:.2f} ms".format(prediction_result.durations_in_seconds["prediction"] * 1000))
 
         if dataset_json_path:
-            if source_is_video == True:
+            if source_is_video is True:
```
cemilcengiz, line 538: Just using `if source_is_video` is enough

sahi/auto_model.py
```diff
@@ -0,0 +1,72 @@
+from typing import Dict, Optional
+
+from sahi.utils.file import import_model_class
+
+MODEL_TYPE_TO_MODEL_CLASS_NAME = {
```
cemilcengiz, line 5: I think this registry should be better as class variable inside AutoDetectionModel. Also, why don't save the classes directly instead of saving only their name?

fcakyon: yes better move it inside the class.

saved only the names to avoid adding a new module import once a new detector model support is added. this way we only need to update a string (instead of adding module import).

can switch to saving directly the classes if its a better practice.

cemilcengiz: Why don't you want to avoid imports? Are they too expensive (I doubt it) or some of them require installation of other packages? As for the better practice, this way you can't catch bugs e.g. incorrectly written class names would not be noticed until the interpreter tries to instantiate them during run time.

PR #470, opened 2022-05-29
refactor automodel

fcakyon:
update `from_layer` metod to be more generic

### DIFF
--- a/sahi/auto_model.py
+++ b/sahi/auto_model.py
@@ -120,7 +120,8 @@ class AutoDetectionModel:
         else:
             raise Exception(f"Unsupported model: {type(layer_model)}. Only YOLOv5 models are supported.")
 
-        DetectionModel = MODEL_TYPE_TO_MODEL_CLASS_NAME[model_type]
+        class_name = MODEL_TYPE_TO_MODEL_CLASS_NAME[model_type]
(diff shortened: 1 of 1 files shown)


sahi/auto_model.py
```diff
@@ -119,16 +116,17 @@ def from_layer(
...
-                category_remapping=category_remapping,
-                image_size=image_size,
-            )
+            model_type = "yolov5"
         else:
             raise Exception(f"Unsupported model: {type(layer_model)}. Only YOLOv5 models are supported.")
 
-        return model
+        DetectionModel = MODEL_TYPE_TO_MODEL_CLASS_NAME[model_type]
+        return DetectionModel(
```
devrimcavusoglu, line 125: MODEL_TYPE_TO_MODEL_CLASS_NAME is a `Dict[str, str]`, so `DetectionModel` is also a `str` object. I think we have a typo here, DetectionModel was first needs to be imported/loaded and then instantiated.

devrimcavusoglu (changes requested):

devrimcavusoglu (approved):

sahi/auto_model.py
fcakyon: fixed :+1:

PR #512, opened 2022-06-25
if none any slicing params calculate automatic these params

mcvarer:
- If I do not want to give slicing parameters or 
- I do not know what to give, 

 it **automatically calculates** slicing parameters according to the resolution of the image.


fcakyon:
@mcvarer this is an awesome contribution!

Would you mind adding a few unittests for some of the functions you introduced? You can create a new file named `tests/test_autoslice.py` and take slicing tests as reference: https://github.com/obss/sahi/blob/main/tests/test_slicing.py

mcvarer:
> @mcvarer this is an awesome contribution!
> 
> Would you mind adding a few unittests for some of the functions you introduced? You can create a new file named `tests/test_autoslice.py` and take slicing tests as reference: https://github.com/obss/sahi/blob/main/tests/test_slicing.py

Thank you so much. @fcakyon

sahi/slicing.py
```diff
@@ -466,3 +469,205 @@ def slice_coco(
...
+
+
+def calc_slice_params(height: int, width: int):
+    """
+    According to Image HxW calculate overlap sliding window and buffer params
+    factor is the power value of 2 closest to the image resolution.
+        factor <= 18: low resolution image such as 300x300, 640x640
+        18 < factor <= 21: medium resolution image such as 1024x1024, 1336x960
+        21 < factor <= 24: upper medium resolution image such as 2048x2048, 2048x4096, 4096x4096
+        factor > 24: high resolution image such as 6380x6380, 4096x8192
```
devrimcavusoglu, line 656: I'm not sure about this as it may not reflect the true standards (but it approximates), but what about changing `upper medium -> high` and `high -> ultra high` ? @mcvarer @fcakyon

sahi/slicing.py
```diff
@@ -466,3 +469,205 @@ def slice_coco(
...
+        slice_row, slice_col, overlap_height_ratio, overlap_width_ratio = slide * 2, slide, ratio, ratio
+    elif orientation == "square":
+        slice_row, slice_col, overlap_height_ratio, overlap_width_ratio = slide, slide, ratio, ratio
+
+    return slice_row, slice_col, overlap_height_ratio, overlap_width_ratio  # noqa
+
+
+def calc_resolution_factor(resolution: int) -> int:
+    """
+    According to image resolution calculate power(2,n) and return most close number return
```
devrimcavusoglu, line 497:
```suggestion
    According to image resolution calculate power(2,n) and return the closest smaller `n`.
```

I added smaller as it acts like floor rather than round.

sahi/slicing.py
```diff
@@ -466,3 +469,205 @@ def slice_coco(
...
+
+    """
+    expo = 0
+    while np.power(2, expo) < resolution:
+        expo += 1
+
+    return expo - 1
+
+
+def cal_aspect_ratio_orientation(width: int, height: int) -> str:
```
devrimcavusoglu, line 511: Do we actually need this function ? Maybe I didn't get it, but why not simply check `height>width`, `height<width`, `height=width` ?

sahi/slicing.py
```diff
@@ -466,3 +469,205 @@ def slice_coco(
...
+    y = int(height / r)
+    if x < y:
+        return "vertical"
+    elif x > y:
+        return "horizontal"
+    else:
+        return "square"
+
+
+def low_resolution(height: int, width: int):
```
devrimcavusoglu, line 537: please combine all of these in one function `low_resolution`, `medium_resolution`, `upper_medium_resolution`, `high_resolution`, and introduce parameter(s) for marginal values e.g `ratio` and `slide`.

sahi/slicing.py
```diff
@@ -30,7 +30,7 @@
 def get_slice_bboxes(
     image_height: int,
     image_width: int,
-    slice_height: int = 512,
+    slice_height: int = None,
```
devrimcavusoglu, line 33: Well, I think there can be better alternative instead of this. My thought is if we gonna make this `None`, make both `slice_height` and `slice_width` `None`, and introduce a new parameter for auto compute, like the following

```python
def get_slice_bboxes(
  ...,
  slice_height: int = None,
  slice_width: int = None,
  auto_slice_resolution: bool = True,
  ...,
)

if slice_height and slice_width:  # If both are not None
  y_overlap = int(overlap_height_ratio * slice_height)
  x_overlap = int(overlap_width_ratio * slice_width)
elif auto_slice_resolution: 
  x_overlap, y_overlap, slice_width, slice_height = calc_slice_params(height=image_height, width=image_width)
else:
  raise ValueError("Compute type is not auto and no values are provided.")  # Mine is just a placeholder message
```

, or alternatively

```python
def get_slice_bboxes(
  ...,
  slice_height: int = 512,
  slice_width: int = 512,
  auto_slice_resolution: bool = False,
  ...,
)

if auto_slice_resolution:  
  x_overlap, y_overlap, slice_width, slice_height = calc_slice_params(height=image_height, width=image_width)
elif slice_height and slice_width:  # If both are not None
  y_overlap = int(overlap_height_ratio * slice_height)
  x_overlap = int(overlap_width_ratio * slice_width)
else:
  raise ValueError("Compute type is not auto and no values are provided.")  # Mine is just a placeholder message
```
In the second implementation, we preserve default values as `512`, but if `auto_slice_resolution` is `True`, then it disregards whatever `slice_height` and `slice_width` is given.

wdyt ? @mcvarer @fcakyon

sahi/slicing.py
```diff
@@ -519,141 +519,94 @@ def cal_aspect_ratio_orientation(width: int, height: int) -> str:
...
     Returns:
-        default overlap params
+        x_overlap, y_overlap, slice_width, slice_height
     """
-    slice_height = height // 1
-    slice_width = width // 1
 
-    x_overlap = slice_width * 1
-    y_overlap = slice_height * 1
+    if resolution == "low":
```
devrimcavusoglu, line 543: Thanks for the update. One slight improvement could be to define `split_row`, `split_col`, `overlap_width_ratio`, `overlap_height_ratio` under if statements, and then after if blocks, perform the following only once, no need to repeat this under each if block.

```python
        slice_height = height // split_col  # noqa
        slice_width = width // split_row  # noqa

        x_overlap = int(slice_width * overlap_width_ratio)  # noqa
        y_overlap = int(slice_height * overlap_height_ratio)  # noqa
```

devrimcavusoglu:
Thanks for the updates @mcvarer. Currently, the PR LGTM. However, I will point one issue, now as this PR changing the default behavior of `slice_height` and `slice_width` to `None` (formerly both were `512` as default) and auto-computes them. Since the default behavior is changed, I think we need to put a warning message for few versions on stating that the default behavior is changed. Should we put such a warning message ? If so, how and where should we put the message ?

My thought is to put such warning under `get_sliced_bboxes()`, wdyt ? @fcakyon

mcvarer:
> Added comments for the last few touches. Thanks for the effort.

Thank you, i have completed your code reviews and will be commit asap after tested. @devrimcavusoglu

sahi/predict.py
```diff
@@ -171,6 +172,9 @@ def get_sliced_prediction(
             The length of buffer for slices to be used during sliced prediction, which is suitable for low memory.
             It may affect the AP if it is specified. The higher the amount, the closer results to the non-buffered.
             scenario. See [the discussion](https://github.com/obss/sahi/pull/445).
+        auto_slice_resolution: bool
+            if not set slice parameters (slice_height, slice_width),
```
devrimcavusoglu, line 176:
```suggestion
            if slice parameters (slice_height, slice_width) are not given,
```

sahi/slicing.py
```diff
@@ -63,14 +64,19 @@ def get_slice_bboxes(
     """
     slice_bboxes = []
     y_max = y_min = 0
+    logger.warning(
+        "Starting from version `0.10.2`, `auto_slice_resolution=True` is introduced as "
+        "the default behavior for determining slice height and width automatically "
+        "calculated by the image properties (resolution, aspect ratio and orientation."
```
devrimcavusoglu, line 70:
```suggestion
        "calculated by the image properties (resolution, aspect ratio and orientation)."
```

devrimcavusoglu (changes requested):

devrimcavusoglu (approved):

devrimcavusoglu:
LGTM :+1:

Issue #526, opened 2022-07-22
Segmentation fault (core dumped)


seekingdeep:
Peace be upon you

When running Sahi
```
(yolo) home@home-lnx:~/programs/yolov5$  sahi predict --source 1.jpg --model_type yolov5 --model_path ./best.pt --slice_height 256 --slice_width 256
Segmentation fault (core dumped)
```

steps to recreate:
```
conda create -n yolo python=3.9
conda activate yolo
pip install sahi
conda install pytorch=1.10.2 torchvision=0.11.3 cudatoolkit=11.3 -c pytorch
pip install yolov5
sahi predict --source 1.jpg --model_type yolov5 --model_path ./best.pt --slice_height 256 --slice_width 256
```

Note:
- python 3.9 i even tried python 3.7
- sahi 0.10.1
- pytorch=1.10.2 torchvision=0.11.3 cudatoolkit=11.3
- yolov5 6.1.6

fcakyon:
Can you please paste here the full error trace?

wgq18:
I met the same problem and find error about 'import torch'. I added 'import torch' in the first line of yolov5_sahi.py and solved this problem. But I don't know how to solve this problem when use 'sahi predict' command. @fcakyon

wgq18:
I want to add sahi inference to Megvii yolox detector or yolov7 detector, can you give me some guide or demo? Please email 1017892298@qq.com, thank you. @seekingdeep

kadirnar:
> I want to add sahi inference to Megvii yolox detector or yolov7 detector, can you give me some guide or demo? Please email [1017892298@qq.com](mailto:1017892298@qq.com), thank you. @seekingdeep

Hi @wgq18 ,

You can run the YOLOX model using the MMDetection library.
YOLOX+ SAHI: https://huggingface.co/spaces/fcakyon/sahi-yolox 

To add the Yolov7 library, simply write the inference code.  I combined the Yolov6 model with SAHI.
https://github.com/kadirnar/Yolov6-SAHI/blob/main/sahi/model.py#L1087-L1135

A pip version will be released for yolov7 soon. After the pip version is out, I will add it to the SAHI library.
Yolov7-Pip:
https://github.com/open-yolo/yolov7/tree/main

PR #541, opened 2022-08-01
fix segmentation fault in same cases

fcakyon:
fixes https://github.com/obss/sahi/issues/526

thanks to @seekingdeep, @wgq18, @anilozfuttu

### DIFF
--- a/sahi/predict.py
+++ b/sahi/predict.py
@@ -32,7 +32,11 @@ from sahi.utils.cv import (
     visualize_object_predictions,
 )
 from sahi.utils.file import Path, increment_path, list_files, save_json, save_pickle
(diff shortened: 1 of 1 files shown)


devrimcavusoglu (approved):
I didn't get why is this happening, but speculating it may be a bug. Although its reported that this change solves the issue, can we also face this on TF ? (in that case we may also need to add that)

fcakyon:
@devrimcavusoglu this seem to be a known Pytorch related bug:

https://github.com/pytorch/vision/issues/4877 https://github.com/pytorch/vision/issues/5919

For now there is no such issue with Tensorflow.

PR #559, opened 2022-08-12
fix an incorrect url in comments

aynursusuz:


fcakyon:
Again, @aynursusuz thanks a lot for the catch!

fcakyon (approved):

Issue #565, opened 2022-08-19
slice imge has error, "Too many open files"


jiangwenj02:
**Traceback**

```python
Traceback (most recent call last):
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/sahi/utils/cv.py", line 156, in read_image_as_pil
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/PIL/Image.py", line 3092, in open
OSError: [Errno 24] Too many open files: 'mmdetection/data/tianyan/images/train/DJI_0591MP416.jpg'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/bin/sahi", line 8, in <module>
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/sahi/cli.py", line 31, in app
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/fire/core.py", line 141, in Fire
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/fire/core.py", line 471, in _Fire
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/fire/core.py", line 681, in _CallAndUpdateTrace
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/sahi/scripts/slice_coco.py", line 59, in slice
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/sahi/slicing.py", line 468, in slice_coco
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/sahi/slicing.py", line 311, in slice_image
  File "/home1/users/jiangwenj02/anaconda3/envs/mmtrack/lib/python3.7/site-packages/sahi/utils/cv.py", line 162, in read_image_as_pil
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 724, in exec_module
[...]

PR #575, opened 2022-08-29
fix a slice_image error

fcakyon:
fixes https://github.com/obss/sahi/issues/565

see https://stackoverflow.com/questions/68785427/oserror-errno-24-too-many-open-files

### DIFF
--- a/sahi/slicing.py
+++ b/sahi/slicing.py
@@ -297,7 +297,7 @@ def slice_image(
         slice_file_path = str(Path(output_dir) / slice_file_name)
         # export sliced image
         image_pil.save(slice_file_path)
-        image_pil.close() # to fix https://github.com/obss/sahi/issues/565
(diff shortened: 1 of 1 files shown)


sahi/slicing.py
```diff
@@ -258,7 +258,6 @@ def slice_image(
     sliced images.
 
     Args:
-        auto_slice_resolution:
```
devrimcavusoglu, line 261: is this mistakenly removed ?

devrimcavusoglu (approved):

sahi/slicing.py
fcakyon: no i deleted it since it is a redundant line

fcakyon: fixed it

devrimcavusoglu (approved):

PR #589, opened 2022-09-16
add unicode language support to save_json()

iwaitu:
save_json() can support unicode language.

### DIFF
--- a/sahi/utils/file.py
+++ b/sahi/utils/file.py
@@ -39,7 +39,7 @@ def save_json(data, save_path):
 
     # export as json
     with open(save_path, "w", encoding="utf-8") as outfile:
-        json.dump(data, outfile, separators=(",", ":"), cls=NumpyEncoder)
+        json.dump(data, outfile, separators=(",", ":"), cls=NumpyEncoder,ensure_ascii=False)
 
 
 # type check when save json files


fcakyon:
Thanks a lot for your contributions!

fcakyon:
@iwaitu can you please reformat your code by following the steps given here: https://github.com/obss/sahi#contributing

sahi/utils/file.py
```diff
@@ -39,7 +39,7 @@ def save_json(data, save_path):
 
     # export as json
     with open(save_path, "w", encoding="utf-8") as outfile:
-        json.dump(data, outfile, separators=(",", ":"), cls=NumpyEncoder)
+        json.dump(data, outfile, separators=(",", ":"), cls=NumpyEncoder,ensure_ascii=False)
```
devrimcavusoglu, line 42: This may not be always wanted, I think it'd be more proper if we add this, ass an argument to `save_json()` function.

devrimcavusoglu: @fcakyon I'm not sure if in sahi we could potentially use `ensure_ascii=False` anywhere, but making it a function argument and let user decide is more generic and better option IMO (to be on safe-side). There's a [discussion in SO](https://stackoverflow.com/q/40412714) about the usage.

fcakyon: yes it seems in most cases `ensure_ascii=False` cause trouble.

PR #599, opened 2022-09-26
Changed Mask class to store RLE encoded masks to save RAM. [WIP]

ChristofferEdlund:
Changed Mask class to store RLE encoded masks to save RAM usage on large images. (Work in progress)


fcakyon:
Wow @ChristofferEdlund this is an amazing contribution!

fcakyon:
@ChristofferEdlund you should follow these steps for proper style formatting:

```bash
pip install sahi[dev]
```

```bash
```

sahi/annotation.py
```diff
@@ -14,6 +13,11 @@
 )
 from sahi.utils.shapely import ShapelyAnnotation
 
+try:
+    from pycocotools import mask as mask_utils
+    use_rle = True
+except:
```
fcakyon, line 19: exception should be defined:

```suggestion
except ImportError:
```

setup.py
```diff
@@ -38,6 +38,7 @@ def get_version():
     extras_require={
         "tests": ["pytest", "mmdet", "norfair", "layer"],
         "dev": ["black==21.7b0", "flake8==3.9.2", "isort==5.9.2", "jupyterlab==3.0.14"],
+        "rle": ["pycocotools"]
```
fcakyon, line 41: we can directly add it to the main requirement.txt file otherwise continuous integration tests wont install this package and wont be able to test your contribution

ChristofferEdlund: Just be aware that doing that might break the sahi build in windows.

fcakyon: latest version of pycocotools works in windows without any issue, we are always using it 👍

fcakyon: thanks to this line: https://github.com/ppwwyyxx/cocoapi/blob/fba5219fb0b3e3c390874178c4acf19624da8bb9/PythonAPI/setup.py#L14

fcakyon:
You should not reformat with the latest version of black, you should you the versions installed with `pip install sahi[dev]`.

black version is important: https://github.com/obss/sahi/blob/5f095d599005976a446bd0489395c0c7a1ae7a83/setup.py#L40

setup.py
ChristofferEdlund: Oh wow, that was about time! =D

This will make Microsoft Visual C++ 14.0 (or greater) an requirement. But that might be acceptable? 
Just did a clean install of pycocotools and it worked.

ChristofferEdlund: Pycocotools is now in requirements.txt

fcakyon:
@ChristofferEdlund ram usage and inference benchmark plots seem really nice. How did you generate them?

Would you like to open a separate PR adding scripts generating such benchmark plots?

sahi/annotation.py
```diff
@@ -14,6 +14,13 @@
 )
 from sahi.utils.shapely import ShapelyAnnotation
 
+try:
+    from pycocotools import mask as mask_utils
+
+    use_rle = True
+except ImportError:
+    use_rle = False
+
```
devrimcavusoglu, line 23: since `pycocotools` is moved to `requirments.txt` as a direct dependency, I think we change this import procedure, I think the following would be sufficient

```python
from pycocotools import mask as mask_utils
``` 

but @ChristofferEdlund got a point that installing pycoctools on Windows is not straightforward, should we make this a soft dependency rather ? @fcakyon

ChristofferEdlund: As @fcakyon pointed out, it seems like one can actually install pycocotools on windows right now. But it is still this hurdle to get over:

https://stackoverflow.com/questions/64261546/how-to-solve-error-microsoft-visual-c-14-0-or-greater-is-required-when-inst

fcakyon: then it might be better to make pycocotools optional, c ompiler might cause issues to some users.

fcakyon: i will handle the required updates 👍

ChristofferEdlund: > 

@fcakyon, maybe.. Those plots where generated by monitoring the RAM usage of a docker container running on internal models and data. Let me think if there is a good way to do something similar with default SAHI and some good open source images.

setup.py
```diff
@@ -36,7 +36,7 @@ def get_version():
     python_requires=">=3.6",
     install_requires=get_requirements(),
     extras_require={
-        "tests": ["pytest", "mmdet", "norfair", "layer"],
+        "tests": ["mmdet==2.25.1", "norfair==1.0.0", "layer", "pycocotools==2.0.5"],
```
fcakyon, line 39:
```suggestion
        "tests": ["mmdet==2.25.1", "norfair==2.0.0", "layer", "pycocotools==2.0.5"],
```

lioneldaniel:
@fcakyon similarly, would you accept a soft dependency to `opencv` in addition to `pycocotools` in [coco.py](https://github.com/obss/sahi/blob/0f3cbcbc2c49f00079be0fe441b844bfcf327997/sahi/utils/coco.py#L132) to support json file with coco format containing RLE annotations?

Here would be my patch:
- add these lines into the import section of [coco.py](https://github.com/obss/sahi/blob/0f3cbcbc2c49f00079be0fe441b844bfcf327997/sahi/utils/coco.py#L5) :
```python
try:
    from pycocotools import mask as mask_utils
    import cv2

    use_rle = True
except ImportError:
    use_rle = False
```
- replace the first lines of method `from_coco_annotation_dict` in [coco.py](https://github.com/obss/sahi/blob/0f3cbcbc2c49f00079be0fe441b844bfcf327997/sahi/utils/coco.py#L117-L135) by the following ones:
```python
    @classmethod
    def from_coco_annotation_dict(cls, annotation_dict: Dict, category_name: Optional[str] = None):
        ""
        Creates CocoAnnotation object from category name and COCO formatted
        annotation dict (with fields "bbox", "segmentation", "category_id").

        Args:
            category_name: str
                Category name of the annotation
            annotation_dict: dict
                COCO formatted annotation dict (with fields "bbox", "segmentation", "category_id")
        ""

        if annotation_dict.__contains__("segmentation") and not isinstance(annotation_dict["segmentation"], list):
            has_rle_segmentation = True
        else:
            has_rle_segmentation = False

        if use_rle and has_rle_segmentation:
            # convert RLE to polygons: depend on opencv and pycocotools
            binary_mask = mask_utils.decode(annotation_dict["segmentation"])
            contours, _ = cv2.findContours(binary_mask, cv2.RETR_TREE, cv2.CHAIN_APPROX_SIMPLE)
            poly_segmentation = []
            for contour in contours:
                if contour.size >= 2*3:  # Valid polygon has at least 3 points
                    poly_segmentation.append(contour.flatten().tolist())
            annotation_dict["segmentation"] = poly_segmentation
            has_rle_segmentation = False
        else:
            logger.warning(
                f"Segmentation annotation for id {annotation_dict['id']} is skipped since RLE segmentation format is not supported without pycocotools and opencv"
            )
```

PR #691, opened 2022-10-24
Fixed a bug in greedynmm func

youngjae-avikus:
If there is only one box, there is a object that cannot be displayed in the box because it does not enter keep_to_merge_list.

### DIFF
--- a/sahi/postprocess/combine.py
+++ b/sahi/postprocess/combine.py
@@ -235,6 +235,7 @@ def greedy_nmm(
 
         # sanity check
         if len(order) == 0:
+            keep_to_merge_list[idx.tolist()] = []
             break
 
         # select coordinates of BBoxes according to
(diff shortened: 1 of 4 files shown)


youngjae-avikus:
Well, there's an error in the test code
When there is only one object in an unmerged object or frame, 
I think that box should be included in the list

fcakyon:
@youngjae-avikus thanks for the catch, you are right on that! let us confirm the fix 👍

youngjae-avikus:
This is how I solved the issue I posted a while ago

https://github.com/obss/sahi/discussions/684

fcakyon:
@youngjae-avikus can you also fix the failing 4 tests?

youngjae-avikus:
@fcakyon sure

fcakyon:
@youngjae-avikus i have updated the tests

youngjae-avikus:
@fcakyon 
It's been delayed because of the departure schedule. Thank you!

PR #802, opened 2022-12-30
Update requirements.txt

dean-wetherby-simplisafe:
Loosen the pinned version of pybboxes dependency to allow for newer versions. The latest pybboxes addresses a poetry package resolution issue but can't be used with this package because of the exact version requirement.


fcakyon:
thanks for your contributions!

PR #833, opened 2023-02-21
Add yolov8 to SAHI

NguyenTheAn:
Add Yolov8DetectionModel class to infer Yolov8 with SAHI

### DIFF
--- a/sahi/auto_model.py
+++ b/sahi/auto_model.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, Optional
 from sahi.utils.file import import_model_class
 
 MODEL_TYPE_TO_MODEL_CLASS_NAME = {
+    "yolov8": "Yolov8DetectionModel",
     "mmdet": "MmdetDetectionModel",
     "yolov5": "Yolov5DetectionModel",
     "detectron2": "Detectron2DetectionModel",
--- a/sahi/models/yolov8.py
+++ b/sahi/models/yolov8.py
@@ -1,0 +1,156 @@
+# OBSS SAHI Tool
+# Code written by AnNT, 2023.
+
+import numpy as np
+import logging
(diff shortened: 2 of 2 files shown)


NguyenTheAn:
Yes, I will do that

NguyenTheAn:
yes, I added it

mayrajeo:
Correct me if I'm wrong, but I think that `perform_inference` should convert the image to BGR before predicting. My reasoning is that when we pass numpy array or PIL image to the model for inference, YOLOv8 uses `LoadPilAndNumpy` for processing them. This class has functions

```
    @staticmethod
    def _single_check(im):
        assert isinstance(im, (Image.Image, np.ndarray)), f'Expected PIL/np.ndarray image type, but got {type(im)}'
        if isinstance(im, Image.Image):
            if im.mode != 'RGB':
                im = im.convert('RGB')
            im = np.asarray(im)[:, :, ::-1]
            im = np.ascontiguousarray(im)  # contiguous
        return im
        
    def _single_preprocess(self, im, auto):
        if self.transforms:
            im = self.transforms(im)  # transforms
        else:
            im = LetterBox(self.imgsz, auto=auto, stride=self.stride)(image=im)
            im = im.transpose((2, 0, 1))[::-1]  # HWC to CHW, BGR to RGB
            im = np.ascontiguousarray(im)  # contiguous
        return im
```

First, the image is checked with `_single_check` which converts the image to BGR **only if its `Image.Image`.** Then, before the model performs inference, `_single_preprocess` converts it from BGR to RGB unless there are transforms to apply (`default=None`). However, since `im` here is a numpy array with channel ordering RGB, `_single_preprocess` converts it to BGR.

Example notebook: https://nbviewer.org/gist/mayrajeo/70ff2f7781bf967daf066e4a9e1743ab

PR #836, opened 2023-02-28
Fix RGB channel ordering for yolov8

mayrajeo:
YOLOv8 expects numpy arrays to have BGR channel ordering instead of RGB. Examples here: https://nbviewer.org/gist/mayrajeo/033e842b7593e58135c7694944905937


mayrajeo:
Fixed the tests to ensure that predictions acquired with `Yolov8DetectionModel.perform_inference` are equal with those acquired with using YOLOv8 model directly

PR #873, opened 2023-05-11
Deprecated interface fix: ultralytics Boxes.boxes is deprecated - replaced with Boxes.data

eVen-gits:
While using `ultralytics` prediction interface, a warining is repeatedly printed:

` >WARNING ⚠️ 'Boxes.boxes' is deprecated. Use 'Boxes.data' instead.`

Replaced the `boxes.boxes` with `boxes.data` to suppress the warning. In my tests, nothing breaks.


eVen-gits:
Hopefully this is OK :) Thanks for all your hard work!

PR #882, opened 2023-05-27
Made yolov8 segmentation available

sweetlhare:
Tested


sweetlhare:
I will correct a couple of points, make tests and update the pull request

PR #896, opened 2023-06-08
OpenVino support for yolov8 object detection

dhaval-zala:
I have added [OpenVino](https://www.intel.com/content/www/us/en/developer/tools/openvino-toolkit/overview.html) for Yolov8 object detection. It will optimise the CPU usage. OpenVino allows to run model in intel CPU with less CPU consumption.  This will create great impact on get_sliced_prediction in terms of cpu utilization.

PR #904, opened 2023-06-23
ONNX runtime inference compatibility

maikimati:
ONNX models compatibility to speed up CPU inference in object detection

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ fire
 terminaltables
 requests
 click
+onnxruntime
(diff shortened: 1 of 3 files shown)


karl-joan:
Hey @maikimati, I have some thoughts regarding this implementation.

- Perhaps it would be wise to create a separate function for the image preprocessing in the case I would like to overwrite it.
- The current resizing doesn't maintain aspect ratio, but the slices needn't be squares. The implementation in [YOLOv8](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/data/augment.py#L538) first resizes the longest side and then pads the remaining space. There's also one pull request open for OpenVINO implementation (#896) with the same resizing scheme.
- It would be nice, if you could provide the `load_model` function a dictionary of options to set up the inference session, including alternative execution provider such as OpenVINO.

fcakyon (changes requested):
code styling errors need to be fixed, tests should be added, a demo notebook should be included :)

PR #940, opened 2023-08-31
RTDETR implementation

edugzlez:
I have included a basic wrapper to use **RTDETR** (which is implemented in the ultralytics library).

RTDETR: https://docs.ultralytics.com/models/rtdetr/

I had to change the behavior of the `device` parameter because it was not responding correctly. In this case, I pass it directly to the ultralytics prediction method.

I have included the demo and the download functions of the base model.


fcakyon:
Amazing contribution @edugzlez ! If you could also include some tests for RTDETR ([you can check the yolov8 tests](https://github.com/obss/sahi/blob/main/tests/test_yolov8model.py)), we would gladly merge your PR 👍

fcakyon (changes requested):
unittests needed

Nisse123:
Very interested in RT-DETR support, please complete this PR

devrimcavusoglu:
@edugzlez Can you add the commits on this PR https://github.com/edugzlez/sahi/pull/1 and we can merge this PR.

edugzlez:
Sorry @devrimcavusoglu for the delay. I've already accepted the PR and refreshed the branch.

devrimcavusoglu:
Thank you @edugzlez, I mentioned 1 more issue in the comment that changing the device behavior in `Yolov8DetectionModel` class could be undesired. Can you move that device change to `RTDETRDetectionModel` and update the PR ? Also, add a comment to that device change line why it is needed. Then, we can merge the PR :+1:

felipetobars:
@edugzlez Hi, how could I start testing ultralytics' SAHI implementation for RTDETR? since it is not yet incorporated into the library when installing it with pip

sahi/models/yolov8.py
```diff
@@ -27,7 +27,6 @@ def load_model(self):
 
         try:
             model = YOLO(self.model_path)
-            model.to(self.device)
```
devrimcavusoglu, line 30: @edugzlez Hi, it's been a while I had concerns about this as RTDETR class should not depend on YOLOv8 for the device placement errors, can you override the method on `RTDetrDetectionModel` class instead and make this change ?

edugzlez: @devrimcavusoglu I'veincluded that line and run the formatting script.

devrimcavusoglu: It still seems that this line `model.to(self.device)` is removed, can u push your changes if didn't ? Model placement seems to be missing.

edugzlez: Sorry, I thought you were referring to including it in `RTDetrDetectionModel.load_model`. I have included it for YOLOV8 and for RTDETR. I don't remember it well, but it seems that when I made the change the first time it failed for some reason.

devrimcavusoglu (approved):

devrimcavusoglu:
@edugzlez Thank you for the amazing contribution. 🎉

felipetobars:
Thank you! @edugzlez

felipetobars:
@edugzlez @devrimcavusoglu Hey, I'm trying to test the Google Colab example for SAHI + RT-DETR, and I'm getting the error AttributeError: 'MLP' object has no attribute 'act'. Could it be a version issue? I installed everything with pip install -U torch sahi ultralytics. A few months ago, I tried it, and this problem didn't occur.

PR #943, opened 2023-09-04
Slicing: add unique slice index number to output file name

jokober:
I had a problem with sahi resulting in wrong bounding boxes after slicing. This is kind of a edge case but I still wanted to propose a solution to that.

This happened due my dataset structure. I have images with the same file name but in different folders. Like so

- File 1: file_name: ./mot-123/img1/000001.jpg
- File 2: file_name: ./mot-456/img1/000001.jpg

After slicing the sliced images get a new name starting with the file name and followed by the slicing coordinates. However sahi does not check if there is already a file available with that name. Hence an existing image file with the same name but an originally different path will be overwritten and the coco dataset will contain several entries linking to the same sliced image but containing different bounding boxes. This resulted in a faulty dataset.

The way I solved it is probably not the prettiest but the fastest I guess. I simply added an index to the slice names resulting in new unique names.


jokober:
I just found this discussion which likely is related: #721

fcakyon (approved):

fcakyon (approved):

fcakyon:
Thanks a lot for your contribution @jokober !

PR #961, opened 2023-11-03
Self-intersection handling

sergiev:
Sometimes you may occur labels that `shapely` considers invalid even if they are generally useful. 
Here's my example: two closest points were placed in such order that in terms of shapely it's self-intersection and therefore attempt to `get_intersection` leads to `shapely.errors.GEOSException`
image
I propose a little workaround for such case: check validity and call `make_valid` if needed.


fcakyon:
@sergiev thanks a lot for your contribution! We can merge this once the tests pass 👍

fcakyon (approved):

PR #988, opened 2024-01-17
Improve yolov8 config

GuillaumeBruand:
- Allow user to pass custom yolo config through the standard yolo yaml config file
- Pass the confidence threshold to inner model to avoid prefiltering with default yolo threshold=0.25

For yolov8, the predict configuration is given at inference time.
We can pass full parameters using the "cfg" argument and a yaml configuration file.

Passing the "conf" argument using confidence_threshold also avoids to prefilter boxes in the underlying yolo model.

### DIFF
--- a/sahi/models/yolov8.py
+++ b/sahi/models/yolov8.py
@@ -59,25 +59,15 @@ class Yolov8DetectionModel(DetectionModel):
         if self.model is None:
             raise ValueError("Model is not loaded, load it by calling .load_model()")
 
-        kwargs = {
-            "cfg": self.config_path,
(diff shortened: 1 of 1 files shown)


fcakyon:
Amazing contribution @GuillaumeBruand ! 🚀 

Please fix the errors raised in tests by:

- Install required development packages:

```bash
pip install -e ."[dev]"
```

- Reformat with black and isort:

```bash
```

Then commit+push

fcakyon:
Thanks for the amazing contribution @GuillaumeBruand !

fcakyon (approved):

PR #1039, opened 2024-05-20
add yolov8 mask support, improve mask processing speed by 4-5x

mayrajeo:
Cleaner version of #888 . Adds support for instance segmentation with yolov8 and makes slicing instance segmentation results use COCO-style annotations. Haven't yet tested with oriented bounding boxes.

Known shortcomings:
* NMS/NMM/GREEDYNMM is still done with bounding boxes instead of polygons
* Not a problem of sahi, but yolov8 models can output multipart polygons even though yolov8 annotation format doesn't support those.


fcakyon (changes requested):

fcakyon:
Amazing job @mayrajeo  🚀

fcakyon (approved):

PR #1066, opened 2024-07-12
Update has_mask method for mmdet models (handle an edge case)

ccomkhj:
Solve the issue mentioned in https://github.com/obss/sahi/discussions/1064, which leads to error when `{with_mask:False}`


fcakyon:
Thanks for your contribution @ccomkhj ! 

Can you please fix the styling so I can merge your PR?

https://github.com/obss/sahi?tab=readme-ov-file#contributing

PR #1094, opened 2024-11-11
empty segmentation fix

timonpalm:
I struggle with an issue that sometimes the model returns an empty mask which eventually let the program fail.

Either poly1 or poly2 was an empty polygon, taking the union with that will also always result in an empty polygon. I simply fixed it by setting the merged polygon to the not empty one.

That worked fine for me and I thought maybe someone else had the same error and this is a reasonable change.


mookghg:
```
    xmin = max(bbox[0], 0)
               ~~~~^^^
TypeError: 'NoneType' object is not subscriptable
```

fcakyon:
Closing this due to inactivity, it might already be fixed in the latest version of SAHI.

PR #1158, opened 2025-05-21
fix: specify a device other than cuda:0

0xf21:
I modified a few lines of code to specify other devices, such as `cuda:1` or `cpu`, instead of only selecting `cuda:0` when instantiating a detection model using the `AutoDetectionModel.from_pretrained` method.


fcakyon:
@0xf21 perfect addition, thanks!

fcakyon (approved):

PR #1194, opened 2025-07-02
Make bounding box immutable

gboeer:
I have picked up on the previous issue about the bounding box immutability.

I propose this change, basically using a frozen dataclass and properties to achieve an immutable class.
I've added some tests to check for immutability.
The existing checks seem to run fine.


fcakyon (approved):
wow perfect job!

PR #1234, opened 2025-08-22
🐛  Update category remapping logic to work with immutable categories

gboeer:
The label remapping logic didn't work with the introduced immutable categories, now it does.
Also added a test for the remapping.

Discussed in:
https://github.com/obss/sahi/discussions/1231#

### DIFF
--- a/sahi/models/base.py
+++ b/sahi/models/base.py
@@ -1,3 +1,4 @@
+import dataclasses
 from typing import Any, Dict, List, Optional
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


sahi/models/base.py
```diff
@@ -1,3 +1,4 @@
+import dataclasses
```
fcakyon, line 1:
```suggestion
```

sahi/models/base.py
```diff
@@ -156,7 +157,10 @@ def _apply_category_remapping(self):
             for object_prediction in object_prediction_list:
                 old_category_id_str = str(object_prediction.category.id)
                 new_category_id_int = self.category_remapping[old_category_id_str]
-                object_prediction.category.id = new_category_id_int
+                object_prediction.category = Category(
+                                    id=new_category_id_int,
+                                    name=object_prediction.category.name
+                                )
```
fcakyon, line 163:
```suggestion
                object_prediction.category = Category(id=new_category_id_int, name=object_prediction.category.name)
```

PR #1245, opened 2025-09-10
Fix currently failing CI by adding a hatch build target

gboeer:
Adding

    [tool.hatch.build.targets.wheel]
    packages = ["sahi"]

to the `pyproject.toml` should fix the current CI errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -162,6 +162,9 @@ requires = ["hatch-mkdocs", "hatch-pip-compile"]
 [tool.hatch.envs.docs]
 detached = false
 
+[tool.hatch.build.targets.wheel]
+packages = ["sahi"]
+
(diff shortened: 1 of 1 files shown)


gboeer:
@fcakyon  Hmm sorry, it seems my change didn't fix the CI. I was misled by the job succeeding in my branch, but I just noticed the that pypi install step was just skipped in my fork. I#ll look into it.

gboeer:
I am now pretty certain the error was introduced here:
https://github.com/obss/sahi/commit/e7109ee3a117d421f64ed383ad4a073b8ff7b000

Especially with the force reinstall from uv removed, I believe that UV is looking for the sahi package at the previously deleted local folder and did not reinstall sahi correctly from pypo anymore.

Still testing ...

PR #1352, opened 2026-04-19
feat: 🚀 allow per-call confidence_threshold override in prediction APIs

onuralpszr:
`AutoDetectionModel.from_pretrained()` has always accepted `confidence_threshold`, but `get_prediction()` and `get_sliced_prediction()` had no way to override it without mutating the model object directly. This made dynamic threshold usage (e.g. sweeping thresholds, per-image tuning) awkward.

Both functions now accept an optional confidence_threshold parameter. When provided, it overrides `detection_model.confidence_threshold` for that call and restores the original value before returning, so the model state is unchanged after the call.

```python
result = get_sliced_prediction(image, model, confidence_threshold=0.7)
# model.confidence_threshold is unchanged after this call
```

Related discussion topic: https://github.com/obss/sahi/discussions/1269

### DIFF
--- a/sahi/predict.py
+++ b/sahi/predict.py
@@ -83,6 +83,7 @@ def get_prediction(
     verbose: int = 0,
     exclude_classes_by_name: list[str] | None = None,
     exclude_classes_by_id: list[int] | None = None,
+    confidence_threshold: float | None = None,
 ) -> PredictionResult:
     """Function for performing prediction for given image using given detection_model.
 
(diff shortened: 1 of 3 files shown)


onuralpszr:
PTAL @fcakyon when you had time. 

sweep conf approach usage

```python
from sahi.predict import get_prediction
model = AutoDetectionModel.from_pretrained(
        model_type="ultralytics",
        model_path="yolo11n.pt",
        confidence_threshold=0.3,
    )
# run with a different threshold without changing the model
result = get_prediction("image.jpg", model, confidence_threshold=0.7)
print(model.confidence_threshold)  # still 0.3
```

fcakyon (approved):

fcakyon:
looks good 👍🏻

PR #1371, opened 2026-06-06
Update and complete Chinese translations for documentation

ZephyrKeXiner:
Adds a complete Simplified Chinese (`zh`) translation of the documentation under `docs/zh/`, served as a standalone localized site at `/zh/` with its own navigation. The English site is unchanged.

1. API reference pages are intentionally kept in English. 

- Their bodies are auto-generated from Python docstrings by mkdocstrings, so translating them would mean forking the source docstrings. 

- Besides, in our experience, Chinese developers tend to prefer the original English for this kind of reference material, where precise wording matters. 

2. Why /zh is built separately (zensical.zh.toml)

- zensical has no per-language nav, so the only way to give the Chinese site its own (Chinese, self-contained) navigation is to build it as a separate site.

- Acturally, due to less experience of zensical, I can't find another way to sync English site and Chinese site.The whole site shares a single nav, and the extra.alternate setting only adds a language-switcher link — it does not generate a localized nav tree.


ZephyrKeXiner:
Sorry, but I didn't modify any code-relate file, only documents. 
I don't know why CI tests all fail.😭

onuralpszr:
@ZephyrKeXiner any special reason you switch "**输出：**" <-- style instead of "## 输出"  basically why 4 star and not h2 block ? I actually changed to h2 but and it was fine ?

ZephyrKeXiner:
@onuralpszr Sorry, my fault sir.

I double-checked the docs history and you are right to question this. The English docs introduced this part as `**Output:**` in b2d8ce3b (#1270), and I could not find a `## Output` version in `docs/cli.md`.

So the mismatch was actually in the Chinese translation: it had used `## 输出：` while the English source used a bold parameter label. I changed it to `**输出：**` to align the Chinese docs with the English docs structure.

I also added a small check script to catch this kind of formatting mismatch between English and Chinese docs in the future.

ZephyrKeXiner:
@onuralpszr I also added a small script (write with Codex, so I can't guarantee its robust. But I will improve it in the progress of maintaining the zh docs) to prevent this kind of issue from happening again.

It compares the Markdown structure of the Chinese docs against the English source docs, including headings, bold label lines, fenced code blocks, admonitions, tables, and `:::` directives. When it finds a mismatch, it prints a clang-style diagnostic(😆) with the Chinese line, the expected English structure, and the source line.

For example:

```bash
python scripts/check_zh_docs_format.py

onuralpszr:
For markdown I was using "https://github.com/rvben/rumdl" <-- and I have local config let me upload this via another PR can you can also test as well. We might dont need script after all plus if we want really pretty table we might also use "prettier" (I can connect to CI/config for local run) so we can prevent another script

onuralpszr:
@ZephyrKeXiner I covered CI errors, please focus on docs as well.

ZephyrKeXiner:
> For markdown I was using "https://github.com/rvben/rumdl" <-- and I have local config let me upload this via another PR can you can also test as well. We might don't need script after all plus if we want really pretty table we might also use "prettier" (I can connect to CI/config for local run) so we can prevent another script

Copy that. So this time I will remove the python script. In the future, I will use rumdl or prettier to manage zh docs.

onuralpszr:
> > For markdown I was using "https://github.com/rvben/rumdl" <-- and I have local config let me upload this via another PR can you can also test as well. We might don't need script after all plus if we want really pretty table we might also use "prettier" (I can connect to CI/config for local run) so we can prevent another script
> 
> Copy that. So this time I will remove the python script. In the future, I will use rumdl or prettier to manage zh docs.

I will add config so you can test as well, before merge let me know if it is works for you (make sure you have backup in case of breakage)

onuralpszr (approved):

PR #1372, opened 2026-06-06
Fix numba parity mismatches with numpy and torchvision

ZephyrKeXiner:
## Summary

Fixes a GreedyNMM parity mismatch in the numba postprocess backend.

The numpy and torchvision GreedyNMM paths preserve score-sorted candidate order when building each keeper's merge list. The numba backend found candidates in score-sorted order, but reconstructed `keep_to_merge` by iterating over original prediction indices, which could produce a different merge order.

Since `_apply_merge()` applies merges sequentially and re-checks matches after each merge, merge order is part of the postprocess semantics.

## Changes

- Return `sorted_idxs` from the numba GreedyNMM inner loop.
- Reconstruct numba `keep_to_merge` using `sorted_idxs` order.
- Add a regression test covering merge-order parity for both IOU and IOS.

## Tests

```bash
PYTHONPATH=. python -m pytest -p no:rerunfailures tests/test_combine.py::TestNumbaBackend -q

### DIFF
--- a/sahi/postprocess/_numba_backend.py
+++ b/sahi/postprocess/_numba_backend.py
@@ -143,10 +143,10 @@ def _greedy_nmm_numba_inner(
     areas: NDArray,
     match_threshold: float,
     use_iou: bool,
-) -> tuple[list[int], NDArray]:
+) -> tuple[list[int], NDArray, NDArray]:
     """Core greedy NMM loop — fully JIT-compiled.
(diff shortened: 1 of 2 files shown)


onuralpszr (approved):

onuralpszr:
@ZephyrKeXiner good catch, thank you very much

PR #1377, opened 2026-06-23
fix: 🐞 enforce dependency and minimum-version checks in import_utils

Ctnn:
## Summary

`check_requirements` and `ensure_package_minimum_version` in`sahi/utils/import_utils.py` ended with a bare `yield`, which silently turned them into **generator functions**. None of their call sites iterate the returned object, so the function bodies never ran and the intended `ImportError` was **never raised**  missing dependencies and too-old package versions passed silently.

## Root cause

A trailing `yield` (left over from a context-manager-style shape that is never actually used with `with`). All affected call sites call the helper without iterating it:

- `sahi/models/base.py` -> `check_requirements(pkgs)` (every model's dependency check)
- `sahi/models/mmdet.py` (module import-time check)
- `sahi/models/huggingface.py` -> `ensure_package_minimum_version("transformers", ...)` (4.42 / 4.49 gates)
- `sahi/predict.py` -> `check_requirements(["fiftyone"])`

Repro before this PR:

```python
from sahi.utils.import_utils import check_requirements
check_requirements(["a_package_that_is_not_installed"])  # returns a generator, raises nothing
```

## Fix

Remove the stray `yield` so both helpers run eagerly and raise as documented.`check_package_minimum_version` already returned a `bool` and is unchanged.

## Impact

Users with a partial install now get a clear, early `ImportError` (e.g. `The following packages are required to use this module: [...]`) instead of a confusing downstream failure during model loading. No behavior change when all dependencies are present.

## Tests

Adds `tests/test_import_utils.py` (the module previously had no coverage):

- regression guards asserting neither helper is a generator function (`inspect.isgeneratorfunction`)
- missing-package → `ImportError`; all-present → passes
- min-version satisfied / too-low / absent / unknown paths for both `ensure_*` and `check_*`

`ruff check`, `ruff format --check`, and `pytest tests/test_import_utils.py` (18 passed) all green.


onuralpszr:
@Ctnn can you share MSE before and after for impact ?

onuralpszr (approved):

PR #1407, opened 2026-07-30
Stream slices from disk instead of decoding the whole image

akshayrb22:
# Stream slices from disk instead of decoding the whole image

## Problem

`get_sliced_prediction()` called `slice_image()`, which decoded the entire image and
materialised every slice before the first inference ran. For a 39266x29140 (1144 MP)
scan that is a 3.2 GB decoded array plus ~5.0 GB of slices, so it raises `MemoryError`
on a 6 GB machine no matter how small `batch_size` is.

## What changed

When `image` is a file path, slices are now read one horizontal row-band at a time, so
peak memory is set by the band, not by the image.

- `SliceImageStream` (`sahi/slicing.py`) computes slice geometry from the file header
  and yields batches of slices lazily via `iter_batches()`. Three band sources:
  libvips (sequential access, the only one whose peak is not the full image),
  in-memory (arrays and Pillow images, which are already decoded), and a whole-image
  decode fallback so nothing regresses when libvips is absent. `is_streaming` reports
  which case an input falls into.
- `read_image_size()` (`sahi/utils/cv.py`) reads dimensions from the header without
  decoding, EXIF quarter-turns included.
- `_read_local_image_as_arr()` decodes local files with OpenCV straight to HWC RGB.
  The old PIL path cost two extra full-size buffers via `Image.tobytes()`.
- `get_sliced_prediction()` consumes `iter_batches()`. Slice export moved inside the
  loop via a new `SliceExporter`, which writes each batch in the background and drains
  before the next one.
- `PredictionResult.image` is a lazily decoded `functools.cached_property`, so a result
  object no longer re-decodes a gigapixel image nobody asked for.
- New optional extra: `pip install sahi[bigimage]`.
- Docs: new sections in `docs/predict.md` and `docs/slicing.md`, mirrored into
  `docs/zh`. The `docs/tr` mirror is left for a Turkish speaker in a follow-up.

Existing public API is unchanged; `slice_image()` still behaves exactly as before.

## Measured

Reproduce everything below with:

```console
pip install -e ".[bigimage]"
python scripts/benchmark_streaming_slices.py                  # slicing alone
python scripts/benchmark_streaming_slices.py --consume-ms 30   # with simulated inference
python scripts/benchmark_streaming_slices.py --image your_scan.jpg
```

Square gaussian-noise JPEGs, 640x640 slices at 0.2 overlap, batch 8, peak RSS of a
fresh process (so the interpreter and imports are in both columns). Ubuntu, libvips
8.15.1, Python 3.13. Medians of 3 runs; memory was near-deterministic, wall time
varied by under 5%:

| image | eager (`slice_image`) | streaming (`SliceImageStream`) |
| --- | --- | --- |
| 1 MP | 0.27 s / 156 MB | 0.44 s / 156 MB |
| 16 MP | 0.62 s / 163 MB | 0.91 s / 189 MB |
| 64 MP | 1.83 s / 438 MB | 2.07 s / 327 MB |
| 144 MP | 3.71 s / 896 MB | 4.13 s / 508 MB |
| 256 MP | 6.56 s / 1537 MB | 6.69 s / 733 MB |

And the case the change exists for — a real 1144 MP JPEG scan, 4389 slices (measured
separately; the sweep above does not go this large):

| path | peak RSS | wall |
| --- | --- | --- |
| streaming (libvips) | 671 MB | 10.9 s |
| `get_sliced_prediction` end to end | 668 MB | 10.9 s |
| whole-image decode alone, no slicing | 6613 MB | 9.0 s |
| `slice_image` (old path) | 3.2 GB + 5.0 GB of slices, `MemoryError` | - |

**Speed depends entirely on whether anything is consuming the slices.**

Slicing alone, streaming is ~10-15% *slower* than eager (the table above), and below
~64 MP it costs memory rather than saving it. But that measures slicing with no model
attached. `iter_batches` now decodes the next band on a worker thread while the caller
works on the current batch - libvips releases the GIL during decode, verified by
cpu-time-over-wall-time (1.97 for two concurrent libvips decodes, 1.00 for the
pure-Python control). At 30 ms of work per batch, about what a small detector costs on
8 slices:

| image | eager | streaming, no prefetch | streaming + prefetch |
| --- | --- | --- | --- |
| 64 MP | 5.77 s / 438 MB | 5.83 s / 327 MB | **4.34 s / 321 MB** |
| 256 MP | 20.73 s / 1537 MB | 19.45 s / 733 MB | **16.62 s / 722 MB** |

So in the path users actually run, streaming is faster than the old eager code *and*
uses half the memory. Prefetch is on by default; `iter_batches(n, prefetch=False)`
turns it off. It is skipped for non-libvips sources, which are already decoded and have
nothing to overlap.

Reviewer note: the prefetch worker owns its libvips source end to end. A `VipsRegion`
is thread-affine — building it on one thread and fetching from another aborts the
process inside libvips — which is why `_open_band_source` is called inside the produced
generator rather than in `iter_batches`, and why `_prefetch` closes that generator on
the producer thread rather than the consumer's. Closing it from the consumer raises
`ValueError: generator already executing` whenever the producer is mid-decode, which
skips the cleanup and strands the worker with a batch and a region pinned.

Memory is not constant in image height either. libvips retains roughly 0.45 bytes per
pixel decoded (~360 MB across the 1144 MP scan), which is the loader's own bookkeeping —
the same climb reproduces with a bare `pyvips.Region` fetch loop and no sahi in the
process. Call it a tenth of a whole-image decode, not a flat line.

## Behaviour changes

- `PredictionResult.image` decodes on first access rather than at construction, so a
  decode error surfaces there instead. A missing file still raises at construction
  (there is a test for it). It stays assignable: `cached_property` writes into
  `__dict__`, so `result.image = img` shadows the cache as the plain attribute did.
  For a numpy input it is no longer a snapshot — mutating the array you passed in now
  shows up in `.image`.
- The `get_sliced_prediction` progress bar counts slices, not batches.
- `durations_in_seconds["slice"]` now covers the time the caller waited for pixels,
  timed around each band read in the loop, plus the geometry step. With prefetching most
  band decoding overlaps inference and is counted in neither, so `["slice"]` is smaller
  than before on the same workload.
- `get_slice_bboxes` now rejects overlap ratios outside `[0.0, 1.0)`. Ratios `>= 1.0`
  already raised; what is new is the rejection of negative ratios, which used to be
  accepted and spaced slices apart, leaving rows no slice covered.
- `SliceImageStream` raises on a zero-size image, matching `slice_image`.
- `read_image_as_pil()` accepts `os.PathLike` as well as `str`, which
  `read_image_size()` already did.
- Images with an EXIF orientation tag, CMYK images, and images with more than 8 bits
  per sample fall back to a whole-image decode, with a warning. libvips converts each
  differently from PIL, and silently wrong pixels are worse than a slower path.
- With `perform_standard_pred=True` (the default) the full image is still decoded once
  for the whole-image pass, which cancels the saving. Leaving it on while slices are
  being streamed now logs a warning naming the flag. Auto-disabling it above some pixel
  threshold would be a silent behaviour change, so it is not done here.
- The `sahi predict` CLI decodes the image up front for the visualisations it exports,
  so it does not benefit from streaming. Documented; use the Python API.

## Tests and CI

`tests/test_slice_stream.py` and `tests/test_predict_stream.py`, 83 tests. Every
streamed slice is compared pixel-for-pixel against `slice_image()` across EXIF
orientations 2-8, grayscale/RGBA/CHW inputs, and image shapes that do not divide evenly
by the slice stride. `tests/test_cv_utils.py` gained a decoder-parity class pinning the
new cv2 fast path against the pre-change PIL semantics across mode/container/EXIF
combinations.

Four of them guard the properties that make this change worth having rather than just
correct:

- `test_read_ahead_stays_bounded` asserts bands are pulled as inference consumes them,
  and that the prefetch worker reads at most one band ahead. The equivalence tests all
  pass against an eager decode, so without this one nothing fails if slicing quietly
  goes back to materialising everything.
- `test_peak_memory_grows_far_slower_than_a_whole_image_decode` streams a 12000x3000
  and a 12000x12000 image in child processes and asserts the peak RSS difference stays
  under a third of the extra decoded bytes. Marked `slow` and deselected in CI, where
  peak-RSS assertions on a shared runner are not dependable; run it with `pytest -m slow`.
- `TestPrefetch` covers what a worker thread gets wrong rather than what pixels it
  produces: a lost exception, reordered batches, and a thread outliving an abandoned
  iteration. One case uses a producer that sleeps *inside* `next()`, since an
  instantaneous producer is always parked in `put` and never reproduces the
  close-from-the-wrong-thread failure described above.
- `TestStandardPredWarning` pins the warning to the streaming case only, so callers who
  passed an already-decoded array are not nagged.

The `streaming` CI job runs two configurations: Ubuntu with system libvips and
`pyvips<3.0` (what `sahi[ci]` and Roboflow `inference` users get, since `inference` caps
pyvips below 3.0 and the libvips-vendoring `binary` extra only exists from 3.0 on), and
the `sahi[bigimage]` extra as documented. The second leg asserts the library really came
from the wheel by importing `_libvips`, the top-level extension module `pyvips-binary`
ships — without that check, an extra that resolved to bindings with no libvips behind
them would silently decode whole images and still pass. **This drops the macOS and
Windows legs an earlier revision had** — the argument is that those tested pyvips' own
wheels rather than sahi, but it is a real coverage reduction and a reasonable thing to
challenge.

`pyvips[binary]` was checked to resolve on both Python 3.8 and 3.12, so the `bigimage`
extra needs no version marker against `requires-python = ">=3.8"`.

## Deliberate choices worth flagging

- **`sahi/utils/cv.py` raises `OPENCV_IO_MAX_IMAGE_PIXELS` at import time.** OpenCV
  reads that variable once when it is imported, so there is no later hook; if something
  imports `cv2` before any `sahi` module, images above 2^30 pixels silently take the
  slower PIL path. Documented in `docs/predict.md` rather than worked around.
- **Batches span row-bands.** Slices are handed out as copies, so carrying the tail of
  one band into the next batch pins nothing but the slices and keeps batches full on
  images only a few slices wide.
- **`_ChunkedBandReader` reads fixed 128-row chunks** rather than fetching each band's
  delta directly. The simpler version quadruples the transient fetch at gigapixel width
  (a 512-row delta is ~60 MB against 15 MB), which is the allocation being bounded.
- **Both band sources check their dimensions against the header read.** libvips refuses
  at open; the in-memory source refuses at construction, on both axes. numpy clamps an
  out-of-range slice instead of raising, so an unchecked disagreement produces silently
  short bands rather than an error.
- **A missing libvips is logged at debug, not warning.** Most installs will not have it
  and its absence is not an error. A libvips that is present but *refuses a file* does
  warn, since the caller may be relying on streaming to fit in RAM.
- **cv2 decode peaks at roughly 2x the final array** (6.6 GB for a 3.2 GB result), so
  the fallback path is not free. That is the ceiling the libvips source removes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,66 @@ jobs:
           # coco analyse
           uv run sahi coco analyse --dataset_json_path tests/data/coco_evaluate/dataset.json --result_json_path tests/data/coco_evaluate/result.json --out_dir tests/data/coco_evaluate/
 
+  streaming:
+    name: "Streaming slice reads (${{ matrix.libvips }} libvips, OS: ${{ matrix.os }})"
+    if:
+      github.event_name != 'schedule' && github.event_name !=
+      'workflow_dispatch'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # System libvips with a pyvips<3.0 pin is what `sahi[ci]`/`inference` users end
+          # up with. See docs/predict.md for why the two cannot be combined.
+          - os: ubuntu-latest
+            libvips: system
+            pyvips: "pyvips<3.0"
+          # The string users are told to type. The leg above hand-installs pyvips, so it
+          # would not notice `[bigimage]` resolving to a pyvips with no libvips behind it,
+          # which falls back to a whole-image decode silently rather than failing.
+          - os: ubuntu-latest
+            libvips: extra
+            pyvips: ""
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Setup uv python package manager
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        with:
+          python-version: "3.12"
+          enable-cache: true
+          activate-environment: true
+          cache-dependency-glob: |
+            **/pyproject.toml
+
+      # -dev rather than libvips42, whose name moved to libvips42t64 in Ubuntu 24.04
+      - name: Install system libvips
+        if: matrix.libvips == 'system'
+        run: sudo apt-get update && sudo apt-get install -y libvips-dev
+
+      # Separate from the `ci` job, whose `inference` dependency caps pyvips<3.0.
+      # Installed without the `binary` extra so streaming is really exercised.
+      - name: Install sahi with libvips
+        if: matrix.libvips != 'extra'
+        run: uv pip install -e . pytest "${{ matrix.pyvips }}"
+      - name: Install sahi with the bigimage extra
+        if: matrix.libvips == 'extra'
+        run: uv pip install -e ".[bigimage]" pytest
+      # The extra must deliver a vendored libvips as well as the pyvips bindings. Without
+      # one, `import pyvips` raises OSError and slicing quietly decodes whole images.
+      # pyvips-binary ships the library as a top-level `_libvips` extension module, not
+      # as a `pyvips_binary` package, so that is what proves the wheel supplied it rather
+      # than a system libvips this leg deliberately does not install.
+      - name: Verify the extra vendored libvips
+        if: matrix.libvips == 'extra'
+        run: python -c "import _libvips; print('vendored libvips at', _libvips.__file__)"
+      - name: Verify libvips is really loadable
+        run: python -c "import pyvips; print('libvips', pyvips.version(0), pyvips.version(1), pyvips.version(2))"
+      # `not slow` drops the peak-RSS measurement: it is a real check but not a dependable
+      # one on a shared runner. See TestPeakMemory and scripts/benchmark_streaming_slices.py.
+      - name: Test streaming slice reads
+        run: pytest tests/test_slice_stream.py tests/test_predict_stream.py -q -m "not slow"
+
   schedule-test:
     name:
       "Schedule/PyPI check (OS: ${{ matrix.os }}, Python: ${{
--- a/README.md
+++ b/README.md
@@ -133,6 +133,16 @@ mim install mmdet==3.3.0
 pip install inference>=0.51.5 rfdetr>=1.6.2
 ```
 
+- Install the streaming backend for gigapixel images (optional, see
+  [slicing docs](docs/slicing.md)):
+
+```console
+pip install sahi[bigimage]
+```
+
+Installing it alongside `inference` needs libvips from your system package manager
+instead — see [installing libvips](docs/predict.md#installing-libvips).
+
 </details>
 
 ## <div align="center">Quick Start</div>
--- a/docs/predict.md
+++ b/docs/predict.md
@@ -39,6 +39,128 @@ result = get_sliced_prediction(
 
 ```
 
+### Predicting on very large images
+
+When you call `get_sliced_prediction` with a file path, slices are read from disk one
+row-band at a time instead of decoding the whole image up front, so peak memory stays a
+small fraction of the scan rather than tracking its full decoded size. Two things unlock
+the full saving:
+
+- Install the optional streaming backend: `pip install sahi[bigimage]` (bundles
+  libvips). Without it the image is decoded whole, matching the old behaviour.
+- Pass `perform_standard_pred=False`. The default also runs inference on the full
+  image, which decodes it in one piece and cancels out the saving. sahi logs a warning
+  when you leave it on while slices are being streamed.
+
+A Pillow image, a numpy array or a URL is already decoded (or has to be fetched whole),
+so those inputs keep the old behaviour. The same goes for the `sahi predict` CLI, which
+decodes the image up front for the visualisations it exports: use the Python API to
+stream.
+
+Images whose EXIF orientation tag rotates or mirrors them cannot be streamed and fall
+back to a whole-image decode (a warning says so). `SliceImageStream.is_streaming` reports
+whether a given input will actually be streamed.
+
+#### Measured effect
+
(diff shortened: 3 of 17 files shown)


onuralpszr:
@akshayrb22 I understand you used LLM for this PR but it is extremely big for me to check as well. I also saw some scripts for testing  and some big tests and you also include new toml optional as well without asking anyone it would be nice to ask as well. It will take some time and I am not quickly jump and merge almost 3k lines in single shot.

akshayrb22:
Hey @onuralpszr, I understand your concerns and criticisms. 

I've converted this to draft. I'll open an issue with the reproduction and numbers: on main, `slice_image()` peaks at ~3.4× the decoded array (measured 2522 MB peak for an image that's 732 MB decoded), so the 1144 MP scan I hit this with needs ~11 GB and OOMs regardless of the `batch_size`. I'll include a synthetic reproduction so no special image is needed to see it.

Then I'll split the work into small PRs with no new dependencies, one at a time, and wait for each to be resolved before opening the next. I'll also drop the benchmark script and trim the test files down to the cases that actually pin behaviour.

Regarding `pyvips` added to `pyproject.toml`, I am happy to table it until the smaller PRs are done and then discuss whether you want it at all. Also, what is your preferred way to handle setting up new tests for any new code introduced? I can keep that in mind while I make my smaller PRs.

akshayrb22:
Following up on this: I have written the measurements up as a discussion — https://github.com/obss/sahi/discussions/1415

It has the peak memory numbers measured against main, where the memory actually goes (Pillow storing RGB at 4 bytes/px, plus `Image.tobytes()` holding the chunk list and the joined copy at the same time), commands to reproduce it, and the benchmark script as a gist.

I will open the first PR shortly. It cuts peak memory by about 38% on its own, with no new dependency and no signature change.

onuralpszr:
Since we start doing smaller PRs and check and discussion I am gonna close this one. I merged other PR :)

PR #1408, opened 2026-07-31
fix: enable MPS auto-detection for postprocessing on macOS

ZephyrKeXiner:
Hello @onuralpszr, 

I found that current version of sahi did not automatically use MPS on macOS, because the code only consider about CUDA device and can't route to MPS automatically. This behavior may be confusing to macOS users and prevents sahi from taking advantage tof the best available device.

This PR adds MPS detection to backend resolution and updates the automatic device selection priority to CUDA → MPS → CPU.

The docs and test part should also be updated if the PR is merged. If you are busy, I can take the part of updating them.😃

---

And by the way, I have one API question: should get_postprocess_backend() return the configured value ("auto") or the concrete backend actually being used? I would expect a getter with this name to report the effective backend, which would help users verify whether hardware acceleration was selected. If both states need to be exposed, we could add a separate getter for the configured value.

### DIFF
--- a/sahi/postprocess/backends.py
+++ b/sahi/postprocess/backends.py
@@ -50,8 +50,8 @@ def resolve_backend() -> str:
 
     When the backend is set to "auto", detection follows this priority:
 
-    1. **torchvision** -- selected if both torchvision and a CUDA GPU are
-       available (GPU-accelerated NMS).
+    1. **torchvision** -- selected if torchvision is installed and a GPU is
+       available, either CUDA or Apple MPS (GPU-accelerated NMS).
     2. **numba** -- selected if the numba package is installed (JIT-compiled
        loops, faster than pure numpy for large prediction counts).
     3. **numpy** -- always available as the fallback (pure numpy,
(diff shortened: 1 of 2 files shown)


onuralpszr:
@ZephyrKeXiner it looks nice and if you can just fix python3.8 CI we can merge as well :=)

onuralpszr:
Thanks for this. I pushed one commit to your branch to get 3.8 green.

`torch.mps.is_available()` only exists from torch 2.5, and `pyproject.toml` pins `torch==2.4.1` for Python 3.8, which is where the 33 failures came from. `torch.backends.mps.is_available()` has been the stable entry point since 1.12 and returns False on non-Apple builds, so it works on both.

I also reverted 8da972a. `torch.has_mps` wasn't what broke 3.8, and it warns on every access — `select_device` is called on every postprocess call through `_torchvision_backend._get_device` — and it is being removed from torch, which would silently drop MPS back to CPU.

Run on an M2 Pro: 227 passed, 11 skipped, with `resolve_backend()` returning torchvision throughout. Python 3.8 with torch 2.4.1 also passes on the files that were failing. NMS timings on the same machine, IOU, best of 5 runs:

| boxes | numpy | numba | torchvision (MPS) |
| ----: | ----: | ----: | ----------------: |
| 100 | 0.14 ms | 0.02 ms | 1.44 ms |
| 500 | 1.53 ms | 0.36 ms | 1.42 ms |
| 1000 | 5.46 ms | 1.40 ms | 1.45 ms |
| 5000 | 81.37 ms | 34.16 ms | 4.05 ms |
| 20000 | 1237.70 ms | 462.56 ms | 11.75 ms |

So your priority change is the right call. numba wins below ~500 boxes but the gap there is about 1.4 ms.

Tests and docs are in #1409 and #1410 so this PR stays small. On your API question: `get_postprocess_backend()` returning `"auto"` is correct as the configured value, and `resolve_backend()` is already the public getter for the effective one. Worth documenting rather than changing.

onuralpszr (approved):
LGTM, Verified on Apple Silicon and on Python 3.8 with torch 2.4.1.

ZephyrKeXiner:
Thank you for your rapid reviewing and CI tests!  And I also learned the stable API of MPS in torch.❤️ 

About the `get_postprocess_backend() ` API, I'm looking forward to latest documents and thank you for your explaining!

PR #1421, opened 2026-08-04
Cut slice_image peak memory by decoding to an array

akshayrb22:
First PR out of the split from #1407.

`slice_image()` uses about 3.4x the memory of the array it slices. Two things in the decode cause it:

- Pillow stores RGB at 4 bytes per pixel, so a 732 MB array decodes into a 976 MB buffer.
- `np.asarray(pil_image)` goes through `Image.tobytes()`, which builds a list of chunks and then joins them, so the pixel data sits in memory twice.

Measured against `2b51db0` (0.12.5) with this branch rebased onto it, so both columns are the same tree apart from these two files. `slice_image()` at 640x640 with 0.2 overlap, each cell in a fresh process, median of three runs:

| input | RGB array | main | this PR | saved |
| --- | --- | --- | --- | --- |
| 121 MP | 346 MB | 1217 MB | 747 MB | 39% |
| 256 MP | 732 MB | 2506 MB | 1520 MB | 39% |

The two columns produce identical slice counts on both inputs (961 at 256 MP). Peak RSS varied by 0 MB across the three repeats, and each run printed the `sahi/slicing.py` it imported so neither column could measure the wrong checkout.

What's in here:

`read_image_size()` reads the dimensions out of the header, so callers that only want a size don't decode the image at all.

`_read_local_image_as_arr()` decodes local files with OpenCV and swaps BGR to RGB in place, which avoids both copies above. It returns None for 16-bit files, unsupported codecs, and anything else OpenCV declines, and the existing PIL path handles those as before. OpenCV is already a required dependency, so this adds nothing to the install.

`slice_image()`'s signature and output don't change. `_slice_file_suffix()` takes over the extension logic that used to read `.filename` off the PIL object we no longer create. The outcome is the same, including exporting png for a path input.

Background and the full set of measurements are in #1415. Let me know if you'd like for me to add tests to this PR as well.


akshayrb22:
I added some new tests to check the changes I made in this PR and rebased on main.

`tests/test_cv_utils.py` checks that:
- `read_image_size()` returns exactly what `read_image_as_pil().size` returns on the images already in the repository.
- The two agree on an image whose EXIF orientation tag turns it a quarter turn - JPEG and TIFF, `exif_fix` on and off. The test compares against the decode instead of hardcoding sizes which makes TIFF work. Pillow's plugin turns it either way while a JPEG turns only when `exif_fix` asks.
- `return_arr=True` gives the same pixels, shape and unit8 dtype as converting from PIL. It is exact for PNG, within 1 for JPEG since two libjpeg builds can round a sample differently.

`tests/test_slicing.py` checks that:
- Slicing a path and slicing an image as an array give identical slices and that a path still exports png.
- `_slice_file_suffix()` returns PNG for a path whatever it's own extension, png for an array, `out_ext` when one is given and a lossless extension such as `.bmp` when an already-open PIL image carries one. The same inputs behave the same way on the commit `2b51db0` so this pins an existing behaviour rather than a change. 

I did not add any new image files. The EXIF, TIFF and BMB cases are written to a temp directory. I also broke the implementation a few ways to ensure that each test actually fails.

onuralpszr (approved):

PR #1422, opened 2026-08-05
Speed up crowded NMM by short-circuiting claimed boxes

ZephyrKeXiner:
## Summary

This PR speeds up the merge bookkeeping in both the stored-CSR and streaming NMM paths.

The existing condition performs a linear membership lookup in `merge_list` before checking whether the candidate has already been claimed:

```python
if m_int not in merge_list and merge_to_keep[m_int] < 0:
```

On crowded inputs, most candidates have already been assigned to a keeper, but the increasingly large Python list is still scanned first.

This PR reverses the two pure conditions:

```Python
if merge_to_keep[m_int] < 0 and m_int not in merge_list:
```

Python can then short-circuit on the O(1) array lookup and avoid the O(group size) list search for claimed candidates. The boolean result is unchanged. 

The same change is applied to nmm_sparse() and nmm_streaming().

## Benchmark

This PR includes `benchmarks/crowded_nmm.py`, a deterministic reproducer for the crowded streaming NMM path.

The benchmark:

- generates crowded boxes from a fixed seed;
- reports the selected NMM route and estimated intersection count;
- validates that every prediction belongs to exactly one output group;
- reports group statistics for correctness comparison;
- optionally profiles the run with `cProfile`.

Input generation and the reporting-only density estimate are outside the timed call. The production route selection performed inside `nmm_numpy()` is included.

### Configuration

- macOS arm64, Apple M4 Pro
- Python 3.11.14
- NumPy 2.4.6
- 10,000 boxes
- IOS, threshold 0.3
- spread 80
- seed 7

Run from the repository root:

```bash
python benchmarks/crowded_nmm.py \
  --boxes 33337 \
  --profile \
  --profile-lines 20
```

### 33,337-box crowded case

| Measurement | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| Total profiled time | 317.284 s | 35.486 s | **8.94x faster** |
| `nmm_streaming()` cumulative time | 316.856 s | 35.071 s | **9.03x faster** |
| `nmm_streaming()` self time | 276.134 s | 6.678 s | **41.35x faster** |
| `MatchQuery.row()` | 38.526 s | 26.361 s | 1.46x faster |
| Native STRtree query | 33.551 s | 22.348 s | 1.50x faster |
| `_dominated_row()` | 2.178 s | 2.016 s | 1.08x faster |
| Row sorting | 1.643 s | 1.555 s | 1.06x faster |
| Function calls | 1,466,571 | 1,466,571 | unchanged |

The total profiled time drops from 317.284 seconds to 35.486 seconds, an 88.8% reduction and an 8.94x end-to-end speedup.

The largest improvement is isolated to the self time of `nmm_streaming()`, which drops from 276.134 seconds to 6.678 seconds. This is a 97.6% reduction in the Python merge bookkeeping time.

The STRtree code is unchanged by this PR. Its timing difference should therefore be treated as run-to-run variation or an indirect cache/runtime effect rather than an STRtree optimization.


ZephyrKeXiner:
I found that every index appended to a merge list is immediately assigned a non-negative entry in `merge_to_keep`, and that entry is never reset.

Therefore, `merge_to_keep[m_int] < 0` guarantees that `m_int` is not already present in any merge list. The linear list membership check is redundant and can be removed.

The current approach is 
```Python
if merge_to_keep[m_int] < 0:
```

